PR #124, opened 2015-01-06
Adding smoothing penalizer to aalen

CamDavidsonPilon:
This PR adds: 

- Added new penalization scheme to AalenAdditiveFitter. You can now add a smoothing penalizer
that will try to keep subsequent values of a hazard curve close together. The penalizing coefficient
is `smoothing_penalizer`. 
- Changed `penalizer` keyword arg to `coef_penalizer` in AalenAdditiveFitter.
- new `ridge_regression` function in `utils.py` to perform linear regression with l2 penalizer terms.

### DIFF
--- a/lifelines/utils.py
+++ b/lifelines/utils.py
@@ -4,6 +4,7 @@ from __future__ import print_function, division
 from datetime import datetime
 
 import numpy as np
+from numpy.linalg import inv
 import pandas as pd
 from pandas import to_datetime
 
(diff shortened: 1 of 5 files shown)


CamDavidsonPilon:
@spacecowboy for review please.

lifelines/utils.py
```diff
@@ -512,3 +513,36 @@ def qth_survival_time(q, survival_function):
 
 def median_survival_times(survival_functions):
     return qth_survival_times(0.5, survival_functions)
+
+
+def ridge_regression(X, Y, c1=0.0, c2=0.0, offset=None):
+    """
```
spacecowboy: I suggest expanding the docs a little. The wikipedia link was really helpful so it should be included in the docstring I think. Perhaps something like:

    Also known as 'Tikhonov regularization'. See http://en.wikipedia.org/wiki/Tikhonov_regularization
    for details. This solves a (possible ill-formed) minimization problem:
         (formula as before)

PR #126, opened 2015-01-29
remove matplotlib from requirements, and making it an optional dep. 

CamDavidsonPilon:
This reduces the dep if you want a cleaner stack. Interestingly enough, Pandas doesn't have mpl as a dependency.

### DIFF
--- a/lifelines/plotting.py
+++ b/lifelines/plotting.py
@@ -1,15 +1,12 @@
 # -*- coding: utf-8 -*-
 from __future__ import print_function
 
-# plotting
 import numpy as np
-import matplotlib as mpl
-from matplotlib import pyplot as plt
-
 from lifelines.utils import coalesce
 
 
 def is_latex_enabled():
+    import matplotlib as mpl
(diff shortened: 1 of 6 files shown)


CamDavidsonPilon:
@spacecowboy any thoughts on this?

lifelines/plotting.py
```diff
@@ -1,15 +1,12 @@
...
-# plotting
 import numpy as np
-import matplotlib as mpl
-from matplotlib import pyplot as plt
-
 from lifelines.utils import coalesce
 
 
 def is_latex_enabled():
+    import matplotlib as mpl
```
spacecowboy: Docstrings have to be the *first* thing in the method, no?
So it should be moved to after the docstring.

spacecowboy:
Seems reasonable otherwise

lifelines/plotting.py
CamDavidsonPilon: :+1:

PR #128, opened 2015-01-31
Adding univariate Weibull to the mix 

CamDavidsonPilon:
This PR adds a WeibullFitter (which uses Newton's Method to find the parameters) and an ExponentialFitter (in another version I'll introduce a PieceWiseExponentialModel). There's also a whole lot of other stuff in this PR - ranging from api changes to cleaning up some code and tests. (lots of the latter).

@spacecowboy, wanna give this a look - sorry the diff is so noisy!

### DIFF
--- a/lifelines/_univariate_weibull.py
+++ b/lifelines/_univariate_weibull.py
@@ -1,5 +1,5 @@
+from __future__ import division
 import numpy as np
-
 
 def _negative_log_likelihood(lambda_rho, T, E):
     if np.any(lambda_rho < 0):
--- a/lifelines/_univariate_weibull.py
+++ b/lifelines/_univariate_weibull.py
@@ -1,0 +1,35 @@
+import numpy as np
+
+
+def _negative_log_likelihood(lambda_rho, T, E):
+    if np.any(lambda_rho < 0):
+        return np.inf
+    lambda_, rho = lambda_rho
+    return - np.log(rho * lambda_) * E.sum() - (rho - 1) * (E * np.log(lambda_ * T)).sum() + ((lambda_ * T) ** rho).sum()
+
+
+def _lambda_gradient(lambda_rho, T, E):
+    lambda_, rho = lambda_rho
+    return - rho * (E / lambda_ - (lambda_ * T) ** rho / lambda_).sum()
+
+
+def _rho_gradient(lambda_rho, T, E):
+    lambda_, rho = lambda_rho
+    return - E.sum() / rho - (np.log(lambda_ * T) * E).sum() + (np.log(lambda_ * T) * (lambda_ * T) ** rho).sum()
+    # - D/p - D Log[m t] + (m t)^p Log[m t]
+
+
+def _d_rho_d_rho(lambda_rho, T, E):
+    lambda_, rho = lambda_rho
+    return (1. / rho ** 2 * E + (np.log(lambda_ * T) ** 2 * (lambda_ * T) ** rho)).sum()
+    # (D/p^2) + (m t)^p Log[m t]^2
+
+
+def _d_lambda_d_lambda_(lambda_rho, T, E):
+    lambda_, rho = lambda_rho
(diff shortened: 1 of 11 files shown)


lifelines/estimation.py
```diff
@@ -29,19 +31,109 @@ def __repr__(self):
         return s
 
 
-class ExponentialFitter(BaseFitter):
+class WeibullFitter(BaseFitter):
+
     """
-    This class implements an exponential model for univariate data. 
+    This class implements an exponential model for univariate data.
```
spacecowboy: Shouldn't this say "...a Weibull model..." ? Or perhaps a sentence on how it differs from the one named "ExponentialFitter"?

spacecowboy: Or the formula below does... Feel free to ignore ;)

CamDavidsonPilon: :+1: nope you're right - copypasta mistake

lifelines/_univariate_weibull.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+
```
spacecowboy: Should add

    from __future__ import division

to avoid differences with python3.

lifelines/_univariate_weibull.py
```diff
@@ -0,0 +1,35 @@
...
+def _d_rho_d_rho(lambda_rho, T, E):
+    lambda_, rho = lambda_rho
+    return (1. / rho ** 2 * E + (np.log(lambda_ * T) ** 2 * (lambda_ * T) ** rho)).sum()
+    # (D/p^2) + (m t)^p Log[m t]^2
+
+
+def _d_lambda_d_lambda_(lambda_rho, T, E):
+    lambda_, rho = lambda_rho
+    return (rho / lambda_ ** 2) * (E + (rho - 1) * (lambda_ * T) ** rho).sum()
```
spacecowboy: Risky division in python2. See top line comment.

lifelines/estimation.py
```diff
@@ -29,19 +31,173 @@ def __repr__(self):
         return s
 
 
-class ExponentialFitter(BaseFitter):
+class WeibullFitter(BaseFitter):
+
     """
-    This class implements an exponential model for univariate data. 
+    This class implements an Weibull model for univariate data. The model has parameterized
```
spacecowboy: "an" -> "a"

lifelines/estimation.py
```diff
@@ -29,19 +31,173 @@ def __repr__(self):
...
-        self.alpha = alpha
+        Returns:
+          self, with new properties like `cumulative_hazard_', 'survival_function_', 'lambda_' and 'rho_'.
+
+        """
+        self.durations = np.asarray(durations, dtype=float)
+        # check for negative or 0 durations - these are not allowed in a weibull model.
+        if np.any(self.durations <= 0):
+            filler = 0.01
+            print("Non-positive times found in durations. Replacing each non-positive value with %.5f" % filler)
```
spacecowboy: I think we should raise a ValueError or something in case of negative durations. A duration can't be negative. If the user is inputting negative durations (as I am), they are not inputting the correct thing.

spacecowboy:
Seems reasonable to me otherwise. I'll assume you've tested the math enough ;)

lifelines/_univariate_weibull.py
CamDavidsonPilon: :+1: good call

lifelines/estimation.py
CamDavidsonPilon: What about zeros? This is tricky as lost of survival analysis has durations of value zero, but users would still want to use weibull.

spacecowboy: What if we raise an error on negative values, but use fill on zeroes then?-

CamDavidsonPilon: Sorry, I mean this model can't even accept 0 as values. According to this model they never happen.

Issue #129, opened 2015-01-31
LogRank statistic for normalized time


spacecowboy:
Doing some logrank tests, I noticed that logrank methods fail for normalized times (where some are negative), which is clearly specified in the documentation:

    event_times_foo: a (nx1) array of event durations (birth to death,...) for the population.

Supporting negative times could be done in two lines (which I do myself before calling running the test atm):

    if np.any(event_times < 0):
        # Be careful not to change the array for the caller
        event_times = event_times - np.min(event_times)

Thoughts?

CamDavidsonPilon:
lgtm! You can send a PR to master or 0.6.0 with this fix

PR #130, opened 2015-02-04
Make logrank test work with negative death times

spacecowboy:
This change is in the spirit of earlier changes to allow normalized
times to be used without issues. Creates birth times as the minimum
death time instead of "zero".

Also changed a bunch of 'max' to 'np.max' to get moar speed.

Associated issue: #129

### DIFF
--- a/lifelines/statistics.py
+++ b/lifelines/statistics.py
@@ -96,8 +96,8 @@ def pairwise_logrank_test(event_durations, groups, event_observed=None,
     if event_observed is None:
         event_observed = np.ones((event_durations.shape[0], 1))
 
-    n = max(event_durations.shape)
-    assert n == max(event_durations.shape) == max(event_observed.shape), "inputs must be of the same length."
+    n = np.max(event_durations.shape)
+    assert n == np.max(event_durations.shape) == np.max(event_observed.shape), "inputs must be of the same length."
     groups, event_durations, event_observed = map(lambda x: pd.Series(np.reshape(x, (n,))), [groups, event_durations, event_observed])
 
     unique_groups = np.unique(groups)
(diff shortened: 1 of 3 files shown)


lifelines/statistics.py
```diff
@@ -208,7 +208,7 @@ def chisq_test(U, degrees_freedom, alpha):
 
 
 def two_sided_z_test(Z, alpha):
-    p_value = 1 - max(stats.norm.cdf(Z), 1 - stats.norm.cdf(Z))
+    p_value = 1 - np.max(stats.norm.cdf(Z), 1 - stats.norm.cdf(Z))
```
CamDavidsonPilon: Investigating this speed up, I see `max` being faster: 
screen shot 2015-02-04 at 08 30 44

CamDavidsonPilon:
Other than the comment, lgtm!

CamDavidsonPilon:
PS will you get a chance to look at this [PR](https://github.com/CamDavidsonPilon/lifelines/pull/128)?

lifelines/statistics.py
spacecowboy: Small arrays are irrelevant as the overhead to go into C with numpy will win out every time.
For large data however:
image

CamDavidsonPilon: ooo console battle! 

But in the cases where we evaluate `max` against `.shape`, the array is only 2 elements.

spacecowboy: In which case a microsecond here or there has no effect. I'll change it back if you prefer.

CamDavidsonPilon: nope leave it - not even sure why I brought it up in the first place :smile:

spacecowboy:
OK, so to clarify, was this good to merge?

CamDavidsonPilon:
:+1:

PR #131, opened 2015-02-04
Math was not entirely correct in C-index

spacecowboy:
A pair which is tied (time_a == time_b) is not an informative pair and
should be ignored.


CamDavidsonPilon:
wow nice catch, :ship:

PR #134, opened 2015-02-26
Allow multiple models in k_fold_cross_validation

spacecowboy:
Often cross_validation is used to compare different models. In such
cases it often desired that all models train on the same data to
make the comparison worthwile.

Scores can now also be anything return by the evaluation function.
One might for example want to compare more than one measure, in which
case a scalar is far too restricting. It is now even possible to just
have the evaulation method dump the validation output and perform
any numerical comparisons later.

Signed-off-by: Jonas Kalderstam <jonas@kalderstam.se>

### DIFF
--- a/lifelines/tests/test_utils.py
+++ b/lifelines/tests/test_utils.py
@@ -231,6 +231,18 @@ def test_cross_validator_returns_k_results():
     assert len(results) == 5
 
 
+def test_cross_validator_returns_fitters_k_results():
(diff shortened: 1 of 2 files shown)


CamDavidsonPilon:
damn, test failures look related to my previous commit. Sorry! I'll fix asap. 

Other than that, this PR looks great! Merge when you're ready.

spacecowboy:
Alrighty!

PR #135, opened 2015-03-01
Version 0.7.0

CamDavidsonPilon:
What started off as a minor patch has turned into a bit more. Changes:

- statistical tests in `lifelines.statstics`. now return a `StatisticalResult` object with properties like `p_value`, `test_results`, and `summary`.  
- removed `suppress_print` kwarg in statistical tests. 
- fixed a bug in how log-rank statistical tests are performed. The covariance matrix was not being correctly calculated. This resulted in slightly different p-values. 
- `WeibullFitter`, `ExponentialFitter`, `KaplanMeierFitter` and `BreslowFlemingHarringtonFitter` all have a `conditional_time_to_event_` property that measures  the median duration remaining until the death event, given survival up until time t.

@spacecowboy for review please!

### DIFF
--- a/lifelines/estimation.py
+++ b/lifelines/estimation.py
@@ -57,7 +57,7 @@ class WeibullFitter(BaseFitter):
         """
         Parameters:
           duration: an array, or pd.Series, of length n -- duration subject was observed for
-          timeline: return the best estimate at the values in timelines (postively increasing)
+          timeline: return the estimate at the values in timeline (postively increasing)
           event_observed: an array, or pd.Series, of length n -- True if the the death was observed, False if the event
              was lost (right-censored). Defaults all True if event_observed==None
           entry: an array, or pd.Series, of length n -- relative time when a subject entered the study. This is
@@ -101,6 +101,10 @@ class WeibullFitter(BaseFitter):
         self.plot_cumulative_hazard = self.plot
 
         return self
+
+    @property
+    def conditional_time_to_event_(self):
+        return _conditional_time_to_event_(self)
 
     def hazard_at_times(self, times):
         return self.lambda_ * self.rho_ * (self.lambda_ * times) ** (self.rho_ - 1)
@@ -245,6 +249,10 @@ class ExponentialFitter(BaseFitter):
 
         return self
 
+    @property
+    def conditional_time_to_event_(self):
+        return _conditional_time_to_event_(self)
+
     def _bounds(self, alpha, ci_labels):
         alpha2 = inv_normal_cdf((1. + alpha) / 2.)
         df = pd.DataFrame(index=self.timeline)
@@ -373,7 +381,7 @@ class NelsonAalenFitter(BaseFitter):
         hazard_name = "differenced-" + cumulative_hazard_name
         hazard_ = self.cumulative_hazard_.diff().fillna(self.cumulative_hazard_.iloc[0])
         C = (hazard_[cumulative_hazard_name] != 0.0).values
-        return pd.DataFrame(1. / (2 * bandwidth) * np.dot(epanechnikov_kernel(timeline[:, None], timeline[C][None, :], bandwidth), hazard_.values[C, :]),
+        return pd.DataFrame(1. / (2 * bandwidth) * np.dot(epanechnikov_kernel(timeline[:, None], timeline[C][None, :], bandwidth), hazard_.values[C,:]),
                             columns=[hazard_name], index=timeline)
 
     def smoothed_hazard_confidence_intervals_(self, bandwidth, hazard_=None):
@@ -469,6 +477,10 @@ class KaplanMeierFitter(BaseFitter):
         setattr(self, "plot_" + estimate_name, self.plot)
         return self
 
+    @property
+    def conditional_time_to_event_(self):
+        return _conditional_time_to_event_(self)
+
     def _bounds(self, cumulative_sq_, alpha, ci_labels):
         # See http://courses.nus.edu.sg/course/stacar/internet/st3242/handouts/notes2.pdf
         alpha2 = inv_normal_cdf((1. + alpha) / 2.)
@@ -490,23 +502,6 @@ class KaplanMeierFitter(BaseFitter):
     def _additive_var(self, population, deaths):
         np.seterr(divide='ignore')
         return (1. * deaths / (population * (population - deaths))).replace([np.inf], 0)
-
-    def conditional_time_to(self):
-        """
-        Return a DataFrame, with index equal to survival_function_, that estimates the median
-        duration remaining until the death event, given survival up until time t. For example, if an
-        indivual exists until age 1, their expected life remaining *given they lived to time 1*
-        might be 9 years.
-
-        Returns:
-            conditional_time_to_: DataFrame, with index equal to survival_function_
-
-        """
-        age = self.survival_function_.index.values[:, None]
-        columns = ['%s - Conditional time remaining to event' % self._label]
-        return pd.DataFrame(qth_survival_times(self.survival_function_[self._label] * 0.5, self.survival_function_).T.sort(ascending=False).values,
-                            index=self.survival_function_.index,
-                            columns=columns) - age
 
 
 class BreslowFlemingHarringtonFitter(BaseFitter):
@@ -569,6 +564,10 @@ class BreslowFlemingHarringtonFitter(BaseFitter):
         self.plot = plot_estimate(self, "survival_function_")
         self.plot_survival_function = self.plot
         return self
+
+    @property
+    def conditional_time_to_event_(self):
+        return _conditional_time_to_event_(self)
 
 
 class AalenAdditiveFitter(BaseFitter):
@@ -1264,7 +1263,7 @@ class CoxPHFitter(BaseFitter):
 
     def _compute_standard_errors(self):
         se = np.sqrt(inv(-self._hessian_).diagonal())
-        return pd.DataFrame(se[None,:],
+        return pd.DataFrame(se[None, :],
                             index=['se'], columns=self.hazards_.columns)
 
     def _compute_z_values(self):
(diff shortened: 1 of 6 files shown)


lifelines/estimation.py
```diff
@@ -1578,10 +1595,6 @@ def qth_survival_time(q, survival_function):
 def median_survival_times(survival_functions):
     return qth_survival_times(0.5, survival_functions)
 
-
-def asymmetric_epanechnikov_kernel(q, x):
-    return (64 * (2 - 4 * q + 6 * q * q - 3 * q ** 3) + 240 * (1 - q) ** 2 * x) / ((1 + q) ** 4 * (19 - 18 * q + 3 * q ** 2))
-
```
CamDavidsonPilon: this was never used, so :fire: it.

lifelines/statistics.py
```diff
@@ -46,14 +45,14 @@ def logrank_test(event_times_A, event_times_B, event_observed_A=None, event_obse
         event_observed_B = np.ones(event_times_B.shape[0])
 
     event_times = np.r_[event_times_A, event_times_B]
-    groups = np.r_[np.zeros(event_times_A.shape[0]), np.ones(event_times_B.shape[0])]
+    groups = np.r_[np.zeros(event_times_A.shape[0], dtype=int), np.ones(event_times_B.shape[0], dtype=int)]
```
CamDavidsonPilon: this was cosmetic. instead of column names being `group:1.0`, etc., I prefer `group:1`.

lifelines/statistics.py
```diff
@@ -157,6 +132,7 @@ def multivariate_logrank_test(event_durations, groups, event_observed=None,
       test_result: True if reject the null, (pendantically) None if we can't reject the null.
 
     """
+    event_durations, groups = np.asarray(event_durations), np.asarray(groups)
```
CamDavidsonPilon: `multivariate_logrank_test` would fail it I put into python lists, so this changes that.

lifelines/statistics.py
```diff
@@ -180,23 +156,56 @@ def multivariate_logrank_test(event_durations, groups, event_observed=None,
...
+    V_ = n_ij.mul(np.sqrt(factor) / n_i, axis='index').fillna(1)
     V = -np.dot(V_.T, V_)
     ix = np.arange(n_groups)
-    V[ix, ix] = V[ix, ix] + ev
+    V[ix, ix] = -V[-1, ix] + V[ix, ix]
+    V = V[:-1, :-1]
 
     # take the first n-1 groups
-    U = Z_j.ix[:-1].dot(np.linalg.pinv(V[:-1, :-1]).dot(Z_j.ix[:-1]))  # Z.T*inv(V)*Z
+    U = Z_j.iloc[:-1].dot(np.linalg.pinv(V[:-1, :-1]).dot(Z_j.iloc[:-1]))  # Z.T*inv(V)*Z
```
CamDavidsonPilon: `iloc` is more defensive than `ix`. If an input duration had `-1`, this value would be in the index, and potentially fail.

lifelines/tests/test_estimation.py
```diff
@@ -227,12 +233,12 @@ def test_weibull_model_does_not_except_negative_or_zero_values(self):
...
     def test_exponential_data_produces_correct_inference_no_censorship(self):
         wf = WeibullFitter()
-        N = 10000
+        N = 40000
         T = 5 * np.random.exponential(1, size=N) ** 2
         wf.fit(T)
         assert abs(wf.rho_ - 0.5) < 0.01
         assert abs(wf.lambda_ - 0.2) < 0.01
-        assert abs(wf.median_ - 5 * np.log(2) ** 2) < 0.01
+        assert abs(wf.median_ - 5 * np.log(2) ** 2) < 0.1  # worse convergence
```
CamDavidsonPilon: This test kept failing on remote tests.

lifelines/estimation.py
```diff
@@ -373,7 +381,7 @@ def smoothed_hazard_(self, bandwidth):
         hazard_name = "differenced-" + cumulative_hazard_name
         hazard_ = self.cumulative_hazard_.diff().fillna(self.cumulative_hazard_.iloc[0])
         C = (hazard_[cumulative_hazard_name] != 0.0).values
-        return pd.DataFrame(1. / (2 * bandwidth) * np.dot(epanechnikov_kernel(timeline[:, None], timeline[C][None, :], bandwidth), hazard_.values[C, :]),
+        return pd.DataFrame(1. / (2 * bandwidth) * np.dot(epanechnikov_kernel(timeline[:, None], timeline[C][None, :], bandwidth), hazard_.values[C,:]),
```
spacecowboy: This is nit-picking, but that space should have been kept :)

lifelines/tests/test_statistics.py
```diff
@@ -1,29 +1,44 @@
...
+    control_E = np.ones_like(control_T)
+
+    treatment_T = [6, 6, 6, 7, 10, 13, 16, 22, 23, 6, 9, 10, 11, 17, 19, 20, 25, 32, 32, 34, 25]
+    treatment_E = [1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+
+    result = stats.logrank_test(control_T, treatment_T, event_observed_A=control_E, event_observed_B=treatment_E)
+    expected_p_value = 4.17e-05
+
+    assert abs(result.p_value - expected_p_value) < 0.0001
+    assert abs(result.test_statistic - 16.8) < 0.1
```
CamDavidsonPilon: This is where I noticed the implementation of logrank was wrong. Previously, lifelines reported p-value was `6.5e-05`  (vs `4.17e-05`) and test statistics 15.931 (vs 16.8).

lifelines/statistics.py
```diff
@@ -180,23 +156,56 @@ def multivariate_logrank_test(event_durations, groups, event_observed=None,
...
+    def __init__(self, test_results, p_value, test_statistic, **kwargs):
+        self.p_value = p_value
+        self.test_statistic = test_statistic
+        self.test_results = test_results
+
+        for kw, value in kwargs.items():
+            setattr(self, kw, value)
+
+        self._kwargs = kwargs
+        self.summary = self.__unicode__()
```
spacecowboy: Should summary perhaps be a property?

    def summary(self):
        return self.__unicode__()

CamDavidsonPilon: Hm, to go along with what we have attached to the `CoxPHFitter`, what about:

```
def print_summary(self):
   return self.__unicode__()

@property
def summary(self):
   cols = ['p-value', 'test_statistics', 'test_results']
   return pd.DataFrame([self.p_value, self.test_statistic, self.test_results], columns=cols)
```

Thoughts?

spacecowboy: Yeah it makes sense to have a unified API in this respect.

lifelines/tests/test_estimation.py
spacecowboy: Kept failing in the probabilistic sense, or consistently?

CamDavidsonPilon: former. I was surprised how slow the sample median converges to the true median.

lifelines/statistics.py
```diff
@@ -180,23 +149,56 @@ def multivariate_logrank_test(event_durations, groups, event_observed=None,
...
+                             alpha=alpha, null_distribution='chi squared',
+                             df=n_groups - 1, **kwargs)
+
+
+class StatisticalResult(object):
+
+    def __init__(self, test_results, p_value, test_statistic, **kwargs):
+        self.p_value = p_value
+        self.test_statistic = test_statistic
+        self.test_results = test_results
```
spacecowboy: Two questions here

1. Given the name "test_results", I am guessing that it can be a list/array/dataframe or results?
2. Should it ever be anything other than boolean(s)? If not, should it perhaps be called "significant"? Or replaced by a method like:

    @property
    def is_significant(self):
        return self.p_value <= self.alpha

?

spacecowboy: It's a bit odd that a result is either "True" or "None".

lifelines/statistics.py
```diff
@@ -180,23 +149,64 @@ def multivariate_logrank_test(event_durations, groups, event_observed=None,
...
+
+        for kw, value in kwargs.items():
+            setattr(self, kw, value)
+
+        self._kwargs = kwargs
+
+    def print_summary(self):
+       print(self.__unicode__())
+       return
```
spacecowboy: Redundant *return* here

lifelines/statistics.py
CamDavidsonPilon: IMO, it's a semantic thing. A test is either reject null (True), or failed to reject null (None, but definitely not False, which would imply the null is True). 

In the latest commit, I've added a `significant` property that is boolean.

CamDavidsonPilon: Actually, I like `is_significant` better.

lifelines/statistics.py
```diff
@@ -180,23 +149,64 @@ def multivariate_logrank_test(event_durations, groups, event_observed=None,
...
+        return "<lifelines.StatisticalResult: \n%s\n>" % self.__unicode__()
+
+    def __unicode__(self):
+        HEADER = "   __ p-value ___|__ test statistic __|__ test results __"
+        meta_data = self._pretty_print_meta_data(self._kwargs)
+        s = ""
+        s += "Results\n"
+        s += meta_data + "\n"
+        s += HEADER + "\n"
+        s += "         %.5f |              %.3f |     %s   " % (self.p_value, self.test_statistic, self.test_results)
```
spacecowboy: Something liket his might produce more consistent output:
```python
# str required on last in case it is None
'{:>10.5f} | {:>10.3f} | {:>10}'.format(p_value, test_statistic, str(test_results))
# Example result
'   0.12346 |    134.123 |       None'
```

lifelines/statistics.py
CamDavidsonPilon: annnnd I'm going to make `test_result` human readable: (Reject or failed to reject)

spacecowboy: OK, makes sense.

lifelines/statistics.py
CamDavidsonPilon: Ah way cleaner

spacecowboy:
Looks good and shiny

spacecowboy:
Something always ****s up the VMs on my machine.... Wheel building is underway.

spacecowboy:
Wheels are done.

http://python-wheels.cowboyprogrammer.org/lifelines/

CamDavidsonPilon:
:+1: uploaded

PR #143, opened 2015-04-06
get the version from version.py

xantares:
this allows to have a __version__ attribute in the module


CamDavidsonPilon:
:+1: I just encountered needing this last week. I'll test locally in an hour or so!

CamDavidsonPilon:
:tophat:'d. Thanks for this contribution!

PR #144, opened 2015-04-17
Add an l2 penalty option to Cox PH model

benkuhn:
Thanks for making such a cool library!

I have some data with a lot of highly correlated attributes that I'd like to fit a Cox model to, so adding an L2 penalty seemed helpful. Maybe it'll be useful to other folks as well.

I'm happy to do more work on this if there's stuff I left out--just let me know!

### DIFF
--- a/lifelines/estimation.py
+++ b/lifelines/estimation.py
@@ -583,10 +583,10 @@ class AalenAdditiveFitter(BaseFitter):
       fit_intercept: If False, do not attach an intercept (column of ones) to the covariate matrix. The
         intercept, b_0(t) acts as a baseline hazard.
       alpha: the level in the confidence intervals.
-      coef_penalizer: Attach a L2 penalizer to the size of the coeffcients during regression. This improves 
-        stability of the estimates and controls for high correlation between covariates. 
+      coef_penalizer: Attach a L2 penalizer to the size of the coeffcients during regression. This improves
+        stability of the estimates and controls for high correlation between covariates.
         For example, this shrinks the absolute value of c_{i,t}. Recommended, even if a small value.
-      smoothing_penalizer: Attach a L2 penalizer to difference between adjacent (over time) coefficents. For 
+      smoothing_penalizer: Attach a L2 penalizer to difference between adjacent (over time) coefficents. For
         example, this shrinks the absolute value of c_{i,t} - c_{i,t+1}.
 
     """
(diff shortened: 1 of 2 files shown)


lifelines/tests/test_estimation.py
```diff
@@ -818,12 +818,31 @@ def test_crossval_for_cox_ph_normalized(self, data_pred2, data_pred1):
...
         df = load_rossi()
         cf = CoxPHFitter(normalize=False)
         cf.fit(df, duration_col='week', event_col='arrest')
         npt.assert_array_almost_equal(cf.hazards_.values, expected, decimal=3)
 
+    def test_penalized_output_against_R(self):
+        # R code:
+        #
+        # rossi <- read.csv('.../lifelines/datasets/rossi.csv')
+        # mod.allison <- coxph(Surv(week, arrest) ~ ridge(fin, age, race, wexp, mar, paro, prio,
```
spacecowboy: I think this is very good information to have here!

spacecowboy:
Cool stuff. Is the default value of the penalizer also 0.0 in R?

benkuhn:
@spacecowboy: Yes, inasmuch as it doesn't have one by default (you add it using the `ridge` function around the relevant features like I did above). But here I set it to zero mainly to preserve compatibility with previous versions.

lifelines/tests/test_estimation.py
CamDavidsonPilon: Agree, thanks!

CamDavidsonPilon:
I like the default zero value here, we can worry about automatic tuning in some future version. 

PR lgtm, thanks for the excellent contribution!

lifelines/estimation.py
```diff
@@ -1152,6 +1158,10 @@ def _newton_rhaphson(self, X, T, E, initial_beta=None, step_size=1.,
                                    include_likelihood=include_likelihood)
             # Do not override hessian and gradient in case of garbage
             h, g = output[:2]
+            if self.penalizer > 0:
+                # add the gradient and hessian of the l2 term
+                g -= self.penalizer * beta.T
```
cosmicBboy: hey @benkuhn thanks for adding this feature! I'm just looking at the code to try to understand L2 reg in the context of cox fitting, and wanted to know why the reg term is `self.penalizer * beta.T` instead of `self.penalizer * beta.T^2`?

CamDavidsonPilon: Good question - why do you think it should be `self.penalizer * beta.T^2`?

cosmicBboy: I most familiar with L2 reg in the context of linear/logistic regression, and typically the formula is
```
min( CostFunction(f(x), y + (Lambda * theta^2)
```
(a la [Wikipedia article](https://en.wikipedia.org/wiki/Regularization_(mathematics)))

But I might be missing something here since I'm not super familiar with the details of fitting a cox model

CamDavidsonPilon: You are right, that is the objective function. However, when you differentiate w.r.t. `beta`, the square term is dropped. That's why it's not present here.

cosmicBboy: ah, that makes sense! thanks for taking the time to explain 👍🏽

PR #146, opened 2015-04-28
Fix setup.py

benkuhn:
Sorry; I said I'd do this a week ago when I messed this up in #145, but then got sidetracked by other stuff at work. Hope the delay wasn't too annoying!


CamDavidsonPilon:
:+1: thanks for following through!

PR #152, opened 2015-05-06
first draft of clogit

CamDavidsonPilon:
@jeffalstott, can you review this and see if I am doing this right? I tried it with a sample dataset, but was getting the wrong numbers. The only difference I can think of is that `CoxPH` uses `efron` for ties, while the R implementation uses the exact method. Thoughts?

### DIFF
--- a/lifelines/estimation.py
+++ b/lifelines/estimation.py
@@ -1442,6 +1442,37 @@ class CoxPHFitter(BaseFitter):
         return baseline_hazard_
 
 
+class CLogit(BaseFitter):
+
+    """
+    This class implements fitting conditional logistic regression:
+
+    Parameters:
+      alpha: the level in the confidence intervals.
+      normalize: substract the mean and divide by standard deviation of each covariate
+        in the input data before performing any fitting.
+      penalizer: Attach a L2 penalizer to the size of the coeffcients during regression. This improves
+        stability of the estimates and controls for high correlation between covariates.
+        For example, this shrinks the absolute value of beta_i. Recommended, even if a small value.
+        The penalty is 1/2 * penalizer * ||beta||^2.
+    """
+
+    def __init__(self, alpha=0.95, normalize=True, penalizer=0.0):
+        self.alpha = alpha
+        self.normalize = normalize
+        assert penalizer >= 0, "penalizer must be >= 0"
+        self.penalizer = penalizer
+
+    def fit(self, X, Y, strata=None):
+        df = pd.DataFrame(X)
+        df['E'] = Y 
+        df['T'] = 1.
+        cp = CoxPHFitter(alpha=self.alpha, penalizer=self.penalizer, normalize=self.normalize)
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
In this gist is a script to play with: https://gist.github.com/CamDavidsonPilon/9fe76773a1895a766ccc

The data comes from http://www.ats.ucla.edu/stat/sas/faq/conditional_logit_sas.htm, which is what I am using to compare my MLE estimates too.

jeffalstott:
I didn't realize you'd be trying this out so soon. Thanks!

To recap:, the survival package in R implements a conditional logit with
clogit. That clogit just calls a coxph, using the 'exact' method. From
coxph's documentation:
""The “exact partial likelihood” is equivalent to a conditional logistic
model, and is appropriate when the times are a small set of discrete
values. If there are a large number of ties and (start, stop) style
survival data the computational time will be excessive."


I started talking with the statsmodels folks, and we determined that
'exact' vs 'efron' is a significant difference. Aside from 'exact' and
'efron' producing different results, the 'exact' method is much faster when
the data is the right shape for a conditional logit. See this thread on the
pystatsmodels Google Group for the conversation:
https://groups.google.com/forum/#!topic/pystatsmodels/o9ZwTPInnuc

Josef at statsmodels implemented a ConditionalLogit over the past two days;
you can see our conversation in the Google Group. Here is the current state
of things, from Josef:

"
I put the module with the class into `miscmodels` which contains
GenericLikelihoodModel subclasses (those are models that are missing
specific bells and whistles)

https://github.com/josef-pkt/statsmodels/commit/5242d49a1cbc2ee8656ea6714d91b7198bf03ba7
<https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2Fjosef-pkt%2Fstatsmodels%2Fcommit%2F5242d49a1cbc2ee8656ea6714d91b7198bf03ba7&sa=D&sntz=1&usg=AFQjCNEoau6iabT8vvxAfxB8vratGcLqrg>

an the script with examples is in a gist for now
https://gist.github.com/josef-pkt/428c3c671c400f5b9cb2
<https://www.google.com/url?q=https%3A%2F%2Fgist.github.com%2Fjosef-pkt%2F428c3c671c400f5b9cb2&sa=D&sntz=1&usg=AFQjCNFALriV2mzzgJcJaU2ZPwEsHSRvTg>
"

Possible next steps would be to collaborate with statsmodels and the
freshly written code there, or to implement ones own 'exact' method. The
statsmodels stuff does need unit tests, etc. I do not know what the license
compatibility is between the two projects.

CamDavidsonPilon:
It's probably more appropriate for it to live in statsmodels - the fact that you can do it in a survival analysis library is just because the likelihoods happen to agree.

I'm going to leave this PR open, and decide later if I want to spend more time on it / even add it to lifelines. I'll be watching the development of the statsmodel version.

PR #162, opened 2015-06-20
Adding recur dataset  + fix failing tests

CamDavidsonPilon:


jayasamraj:
Hi,
  **recur dataset.**
As it is a counting process approach(recurrent event dataset), we need to pass the argument (TIME0 and TIME1)

cph = CoxPHFitter()
cph.fit(df, 'TIME', event_col='E')

Here, we cant pass the 2 arguments as( cph.fit(df, 'TIME0','TIME1' , event_col='E')) in cox.
Could you please help me out for how to do that.

Thanks.

j-greer:
Hi @CamDavidsonPilon,
Does lifeline have any functionality for more modelling data of the same form as 'recur.csv' in the datasets directory?

I noticed some commits but I couldn't find anything explicit in the documentation.

Thanks

Jon

CamDavidsonPilon:
So, there is not formal support for recurring survival data, buuut I think all the pieces are there for a first approximation. With the new addition of `cluster_col` in `CoxPHFitter.fit`, we can do something like:

```python
# untested...
from lifelines import CoxPHFitter
from lifelines.datasets import load_recur

recur = load_recur()

recur['T'] = recur['TIME1'] - recur['TIME0']
recur = recur.drop(['TIME1', 'TIME0'], axis=1)

cph = CoxPHFitter()
# with the `cluster_col` we specify that we expect rows with the same id to have correlations
cph.fit(recur, duration_col='T', event_col='CENSOR', cluster_col='ID')

cph.print_summary()
```

Note that this uses the `cluster_col` which tells the fitter two things: 1) the subjects with the same rows have correlated errors, 2) to use the robust-variance (sandwich) estimator. What is still lacks though is late-entry.

PR #175, opened 2015-07-13
Add at risk

CamDavidsonPilon:
This adds an at-risk column to the survival table


CamDavidsonPilon:
Merging on :red_circle: - the test will be updated in 0.8.0

PR #181, opened 2015-08-04
Update travis config to run on faster containers

spacecowboy:
Also updates Conda urls to something that might be more future proof.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
+# This allows it to run on faster containers.
+sudo: false
+
 language: python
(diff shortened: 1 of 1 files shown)


spacecowboy:
Besides being faster, the builds also start within 10 seconds.

See http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

CamDavidsonPilon:
sweet! lgtm!

PR #190, opened 2015-09-30
Making at_risk and deaths by time period available after KM is fit

jstoxrocky:
I find it helpful to cut out time periods with low population sizes when fitting shifted-beta geometric models to KM curves. Making the number of observants still active by time period available allows me to truncate the KM curve at the appropriate cutoff.

Big fan of the package!!


CamDavidsonPilon:
Hey @jstoxrocky, is the `event_table` property what you are looking for? It accessible as `kmf.event_table`

jstoxrocky:
ah perfect, 
All I needed. Thanks

CamDavidsonPilon:
I see the value in your PR, however I'm hesitant to add more to the API, especially when it's available elsewhere. Hope that's understandable!

Issue #191, opened 2015-10-03
"color_cycle" does not exist in matplotlib 1.5


thomas-haslwanter:
The matplotlib upgrade from 1.4.3 to 1.5.0 breaks lifelines:
In "plotting.py", lines 317/8:
```
        kwargs['color'] = coalesce(kwargs.get('c'), kwargs.get('color'),                                  next(kwargs["ax"]._get_lines.color_cycle))
```
has to be replaced by
```
  import matplotlib as mpl
  if int(mpl.__version__.split('.')[1]) > 4:
        kwargs['color'] = coalesce(kwargs.get('c'), kwargs.get('color'), next(kwargs["ax"]._get_lines.prop_cycler))
```

CamDavidsonPilon:
Nice, thanks for reporting this. I'll get a fix in asap

CamDavidsonPilon:
Uh, maybe you can help - how do I install the latest development version of matplotlib?

thomas-haslwanter:
If you work on Windows, copy the appropriate file
   matplotlib‑1.5.0rc2‑cp35‑none‑win32.whl
or
   matplotlib‑1.5.0rc2‑cp35‑none‑win_amd64.whl
from
   http://www.lfd.uci.edu/~gohlke/pythonlibs/#matplotlib
to your computer, and install it with 
  pip install [filename].

If you are on a Mac, I assume that 
   pip install --pre -f http://wheels.scipy.org matplotlib 
would work (taken from http://matplotlib.1069221.n5.nabble.com/Matplotlib-devel-v1-5-0rc2-td46266.html)

And if you are on Linux, then I don't really know off-hand ....

CamDavidsonPilon:
Closed in ce7e54713bcb01877bd609ca554c5445dfb15c16

thomas-haslwanter:
Cam, could you please update lifelines on pypi?
The currently installed version does not contain the fix that you implemented for this issue.
Thanks, thomas

CamDavidsonPilon:
@thomas-haslwanter, thanks for the notice! Done now: https://pypi.python.org/pypi/lifelines

cleemesser:
I think this fix does not work for matplotlib 2.0, I'm guessing that you are conditioning on the minor version only in your test.

CamDavidsonPilon:
[...]

PR #192, opened 2015-10-03
Enable plotting tests on TravisCI

spacecowboy:
In response to #191, this PR allows the plotting tests to run on Travis. It tests all python versions (added 3.5) with both 1.4.3 and 1.5.0 of matplotlib.

Once 1.5.0 is released for real, the .travis file can be simplified. Atm it installs 1.5.0 with pip (requires extra apt-packages) and 1.4.3 via conda (quick and easy).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,17 @@ language: python
 matrix:
   include:
     - python: 2.7
-      env: MATPLOTLIB_VERSION=v1.4.3
+      env: MATPLOTLIB_VERSION=1.4.3
     - python: 3.3
-      env: MATPLOTLIB_VERSION=v1.4.3
+      env: MATPLOTLIB_VERSION=1.4.3
     - python: 3.4
-      env: MATPLOTLIB_VERSION=v1.4.3
+      env: MATPLOTLIB_VERSION=1.4.3
     - python: 3.5
-      env: MATPLOTLIB_VERSION=v1.4.3
+      env: MATPLOTLIB_VERSION=1.4.3
     - python: 2.7
-      env: MATPLOTLIB_VERSION=v1.5.0rc2
+      env: MATPLOTLIB_VERSION=1.5.0rc2
     - python: 3.5
-      env: MATPLOTLIB_VERSION=v1.5.0rc2
+      env: MATPLOTLIB_VERSION=1.5.0rc2
 #  allow_failures:
 #    - python: "nightly"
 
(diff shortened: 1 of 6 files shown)


spacecowboy:
Note that an error might already have been detected by the tests in: https://travis-ci.org/CamDavidsonPilon/lifelines/jobs/83497588

CamDavidsonPilon:
:+1: thanks for looking at this. I'm confused about what the new `block` parameter does - can you explain it?

spacecowboy:
To run the plotting tests on CI, `plt.show()` must be non-blocking. The way they worked before (and still default behavior) is blocking, e.g. the test suite will stop until you manually close the plot.

Running non-blocking is not useful for a human, as ~18 windows will be created in rapid succession, and they will all be closed when the test ends. So having the old behavior as default is sensible.

CamDavidsonPilon:
right, that makes sense. I see there is a test failure, I've seen this test fail a few times before (so I don't think it is 1.5 related). I'll dig in if you don't have time

spacecowboy:
@CamDavidsonPilon created another branch to investigate it on: https://github.com/CamDavidsonPilon/lifelines/tree/fail-investigate

Printed another error in the same place:https://travis-ci.org/CamDavidsonPilon/lifelines/jobs/83536065

dimension of x, y1 and y2 do not match for some reason

spacecowboy:
@CamDavidsonPilon Tested failed when `shaded_plot` got the following argument shapes:

```
x (1263,)
y (1263, 2)
y_lower (1263, 2)
y_upper (1263, 2)
```

Shape of `x` does not match y-parameters. Strange that it only happens sometimes.

CamDavidsonPilon:
cool I found the cause of that bug, I'm going to commit here.

CamDavidsonPilon:
Here's what was happening. `https://github.com/CamDavidsonPilon/lifelines/blob/master/lifelines/plotting.py#L256`, which `shaded_plot` uses, was sometimes returning >1-d array. This was odd, as they should be unique (or so I thought). Turns out the artificially generated test data can have multiple columns the same: 

screen shot 2015-10-06 at 23 20 38

This columns come from https://github.com/CamDavidsonPilon/lifelines/blob/master/lifelines/generate_datasets.py#L167, and `f.__doc__` is from https://github.com/CamDavidsonPilon/lifelines/blob/master/lifelines/generate_datasets.py#L45. If the random data was generated with a few constant columns + all being within two decimals, they would cause the same column names.

CamDavidsonPilon:
PR looks good to me, :ship: when you're ready!

spacecowboy:
Awesome. Will take another quick look before I merge it.

Issue #147, opened 2015-04-28
predict_median is sometimes a poor default for k_fold_cross_validation


benkuhn:
Currently when one calls `k_fold_cross_validation` on a Cox model, the concordance is calculated using the `predict_median`. For problems where the majority of cases survive the entire study duration, this can be problematic, as many cases end up having median survival time of `inf`. (For a flavor of how severe this is, it knocked a full 0.10 off the concordance index for a problem I'm working on, which really threw me for a loop until I did my own split-sample testing.)

I guess it's hard to find a default here that works well, but using `predict_median` for a Cox model leads to some pretty strange behavior. Here are a couple of solutions I can think of (I'd be happy to implement one but want to run it by others first):

- Default to `predict_partial_hazard` (or rather, its negative) for the Cox model and `predict_median` otherwise
- Allow each model to implement a `predict_rank` function and use that by default
- Switch from `predict_median` to something less likely to generate `inf`s (perhaps `predict_expectation`? not sure how well this would work for non-Cox models)
- Have `k_fold_cross_validation` issue a warning if it seems like there are too many ties in the output of `predict_median`

Which one seems best?

CamDavidsonPilon:
I've encountered this before, one thing to try is also 
```
p = 0.8  # some value larger than 0.5
k_fold_cross_validation(fitters, df, duration_col, event_col,
                            predictor="predict_percentile", predictor_kwargs={'p': p})
```

Though `predict_partial_hazard` is another good suggestion (which is really what these survival models are trying to predict).

CamDavidsonPilon:
If we were to select a new one, I would choose `predict_partial_hazard`. But this lacks interpretability. I guess it's a decision between what performance metric one is using: hazard for concordance, and median/percentile for l1/l2 loss against the actual survival time. (This has it's own issues: it's not totally clear how to measure predictions against censored cases).

benkuhn:
I think that for a Cox model, any prediction function that doesn't introduce extra ties will rank things identically to `predict_partial_hazard`, so there's not really an issue of interpretability there. On the other hand, as I understand it the Aalen additive model doesn't have any similar canonical univariate thing to rank stuff by. I don't think Harrell was really thinking of such non-parametric models when he proposed the concordance index for ranking.

[...]

PR #218, opened 2015-12-26
Remove predict method

CamDavidsonPilon:
I'd like some opinions on this, specifically the change for the `AalenAdditiveFitter.predict` method. @benkuhn and @spacecowboy can you please take a look?

Related to #147.

### DIFF
--- a/tests/test_estimation.py
+++ b/tests/test_estimation.py
@@ -565,7 +565,7 @@ class TestRegressionFitters():
         aaf.fit(X, duration_col='T', event_col='E')
         cph.fit(X, duration_col='T', event_col='E')
 
-        for fit_method in ['predict_percentile', 'predict_median', 'predict_expectation', 'predict_survival_function', 'predict', 'predict_cumulative_hazard']:
+        for fit_method in ['predict_percentile', 'predict_median', 'predict_expectation', 'predict_survival_function', 'predict_cumulative_hazard']:
             assert isinstance(getattr(aaf, fit_method)(X), type(getattr(cph, fit_method)(X)))
 
     def test_duration_vector_can_be_normalized(self):
@@ -588,13 +588,11 @@ class TestRegressionFitters():
         cph.fit(data_pred2, duration_col='t', event_col='E')
         npt.assert_array_equal(cph.predict_partial_hazard(x).index, expected_index)
         npt.assert_array_equal(cph.predict_percentile(x).index, expected_index)
-        npt.assert_array_equal(cph.predict(x).index, expected_index)
         npt.assert_array_equal(cph.predict_expectation(x).index, expected_index)
 
         aaf = AalenAdditiveFitter()
         aaf.fit(data_pred2, duration_col='t', event_col='E')
         npt.assert_array_equal(aaf.predict_percentile(x).index, expected_index)
-        npt.assert_array_equal(aaf.predict(x).index, expected_index)
         npt.assert_array_equal(aaf.predict_expectation(x).index, expected_index)
 
 
@@ -1145,7 +1143,7 @@ class TestAalenAdditiveFitter():
 
             expected = 0.90
             msg = "Expected min-mean c-index {:.2f} < {:.2f}"
-            assert np.mean(mean_scores) > expected, msg.format(expected, scores.mean())
+            assert np.mean(mean_scores) > expected, msg.format(expected, np.mean(scores))
 
     def test_predict_cumulative_hazard_inputs(self, data_pred1):
         aaf = AalenAdditiveFitter()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 #### Forthcoming 0.9.0
  - new prediction function in `CoxPHFitter`, `predict_log_hazard_relative_to_mean`, that mimics what R's `predict.coxph` does.
- - changing the default `predict` method in lifelines to return _not the median_, but another value dependent on the fitter that is calling it. This is because too often the `predict_median` function was returning inf values which would significantly damage measure of concordence index. 
+ - removing the `predict` method in CoxPHFitter and AalenAdditiveFitter. This is because the choice of `predict_median` as a default was causing too much confusion, and no other natual choice as a default was available. All other `predict_` methods remain. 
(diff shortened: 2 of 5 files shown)


tests/test_estimation.py
```diff
@@ -1145,7 +1145,7 @@ def test_crossval_for_aalen_add(self, data_pred2, data_pred1):
 
             expected = 0.90
             msg = "Expected min-mean c-index {:.2f} < {:.2f}"
-            assert np.mean(mean_scores) > expected, msg.format(expected, scores.mean())
+            assert np.mean(mean_scores) > expected, msg.format(expected, np.mean(scores))
```
CamDavidsonPilon: this change was a bug uncovered when this `assert` started failing. 

It's interesting the understand why this test started failing. Initially, for AAF, I used `self.predict_cumulative_hazard(X).iloc[-1].to_frame()` (no negative sign). This means that higher values would mean higher chances of dieing (as it's a hazard). But this test failed because the concordance was backwards - I would rank individuals who have a predicted small lifetimes as having the highest. So I added the negative sign and everything works. 

The problem is now in the interpretation of `predict` for use outside concordance index metrics. The user will get back negative values, which is kind of odd.

benkuhn:
Huh, returning the negative cumulative hazard seems like it would be kind of confusing. On the other hand, I can't think of a particularly simple and natural object to return as the result of `predict()` with no parameters for an Aalen additive model. What was the purpose of this change?

[Also, I believe the documentation for the Aalen predict function ("The higher the [negative cumulative hazard], the less likely to die") is imprecise for discrete-time models. For instance, a discrete 20-period hazard of [1, 0, 0, 0, .....] (negative cumulative hazard of -1) has a much higher likelihood of death than [0.1, 0.1, ....] (negative cumulative hazard of -2). This probably doesn't matter much for real life datasets, but it makes me a bit wary of using cumulative hazard too much for discrete problems...]

IvanBarrientos:
*My two cents*

*Prediction*
From my initial foray into survival analysis, making (lifetime) predictions
is sometimes a red herring in the literature. Books such as Grambsch +
Therneau and Hosmer, Lemeshow, + May shy away from (lifetime) predictions
and emphasize that (standard) survival analysis is focused on the
descriptive rather than predictive.

I, for one, like taking the area under the survival curve (np.trapz) since
this calculation incorporates both the timeline and each respective
probability. On the other hand, the time t' such that median(S(t)) = S(t')
will generally reveal less information about S(t).

In practice, the above area "expectation" can be used as a baseline and
then the practitioner can modify it.

*Concordance*
Is there a good reference for a 'standard' c-index for Aalen models? The
c-index for CoxPH is intuitive in the sense that "risk" := exp(x.\beta)
whereas for Aalen there is not one \beta but a time-varying family of them.
http://www.ncbi.nlm.nih.gov/pubmed/23104317
http://publicifsv.sund.ku.dk/~chpi/talks/CPH-stat-2012/ocspring2012.pdf

Again, fantastic work on Lifelines!

CamDavidsonPilon:
Thanks for the discussion, @benkuhn and @IvanBarrientos.

> What was the purpose of this change?

ATM the default predict is predict_median, which has a tendency to show way too many `infs`. This is a major problem for metrics like the concordance index, which rely on sortable values. This can be solved by returning a hazard for an individual, as this is a (typically unique) value between [0, inf). 

> I can't think of a particularly simple and natural object to return as the result of predict()

Agree, and one solution might be to remove the `predict` method altogether. The consequences of this are we then require `k_fold_cross_validation` to accept a mandatory prediction method, and we are offloading the problem on to packages like https://github.com/sashaostr/sklearn-lifelines. 

> I, for one, like taking the area under the survival curve

That is a good idea, it: 
1) is interpretable (kinda, it's like a truncated expected value)
2) the higher it is, the longer the user should live for, so fits nicely with concordence index. 
3) If we take it equal to the expected value, has nice mathematical properties like minimizing the squared-error loss. 

However
1) it's not a common survival analysis computed stat (hazards are more common, or the full expected value)
2) For the CoxPHFitter, it relies on estimating the baseline hazard, which is difficult to estimate and increases uncertainty. 


> Is there a good reference for a 'standard' c-index for Aalen models? The
c-index for CoxPH is intuitive in the sense that "risk" := exp(x.\beta)
whereas for Aalen there is not one \beta but a time-varying family of them.

Sorry, I don't understand this. Can you explain more?

benkuhn:
> Agree, and one solution might be to remove the predict method altogether. The consequences of this are we then require k_fold_cross_validation to accept a mandatory prediction method, and we are offloading the problem on to packages like https://github.com/sashaostr/sklearn-lifelines.

I have to say I'm a fan of this, or something like it. [Explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/) anyway :)

To me, `predict()` intuitively means something like "give me the prediction that most people want from this particular model most of the time" (at least if you're borrowing its meaning in R). But that requires the output type of `predict` to be almost completely unconstrained, e.g. the Aalen model's natural prediction is multidimensional and the Cox model's natural prediction isn't even probabilities, only relative rankings.

CamDavidsonPilon:
Interestingly, that linked package has already chosen a method: https://github.com/sashaostr/sklearn-lifelines/blob/master/sklearn_lifelines/estimators_wrappers.py#L69

Okay, I think I'm going to remove them. Let me PR something

IvanBarrientos:
Nice discussion and good followups.

Regarding c-index, which I'm specializing to "concordance" in CoxPH: if I
understood correctly once the vector of trained weights, beta in R^n, is
fixed then the "risk" of a member with value x in R^n is the value
exp(x.beta) in R. This risk is not a function of time. When computing
concordance on pairs, we compare pairs of members' respective observed
event times with this risk.

On the other hand, in the Aalen model beta is not a point in R^n but rather
beta(t): T -> R^n where the domain T is a subset of of the non-negative
reals R^\geq 0. Therefore, exp(x.beta(t)) depends on time and we don't
(immediately) get a single"risk" score as in CoxPH.

Does that makes sense? In nutshell: what's the appropriate generalization
of "risk" if beta(t) depends on time?

CamDavidsonPilon:
@IvanBarrientos ah I see your point. Yes, that is an unresolved questions AFAIK

PR #222, opened 2016-01-21
Use setuptools

stevenmanton:
It seems like setuptools is favored over distutils and allows things like installing in develop mode, which is handing when developing/debugging a package (I use that ability frequently). Setuptools isn't in the standard Python distro, but this package already has a ton of other dependencies; it seems reasonable to assume that setuptools is installed in whatever environment lifelines is deployed.

Thanks for an awesome project! I'm working on calculating lifetimes and have already found it very valuable (even just the documentation). I'm looking forward to contributing more.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 *.pyc
 *.so
+build
+dist
+*.egg-info
+*.iml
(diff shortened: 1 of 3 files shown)


CamDavidsonPilon:
oh wow, didn't realize this. Thanks for the heads up and the PR @stevenmanton! Merging now!

stevenmanton:
Me neither, I had to read about it after `python setup.py debug` failed haha

PR #232, opened 2016-04-15
Adds the SBG package to lifelines.

fmfn:
This PR is still a work in progress, but I decided to create it to keep track of progress and get feedback on whatever could use changes.

The goal of this PR is to integrate the Shifted Beta Geometric model into lifelines, and to do so with a consistent API as well as out-of-the-box tools for retention, churn and LTV calculations.

It include the package basic modules as well as a data generating script and an example notebook.


lifelines/utils/DataHandler.py
```diff
@@ -0,0 +1,500 @@
+import numpy as np
+
+
+class DataHandler(object):
```
CamDavidsonPilon: Should this live in lifelines? I wonder if this library should perform data transformations, or we let the user do any preprocessing before invoking lifelines (In my notes, I often rely on patsy/pandas/external libs).

lifelines/utils/ShiftedBetaGeometric.py
```diff
@@ -0,0 +1,726 @@
...
+        alpha = np.clip(alpha, 1e-5, 1e5)
+        beta = np.clip(beta, 1e-5, 1e5)
+
+        # Use death probability of the most recent period
+        p = betaln(alpha + 1, beta + num_periods - 1)
+        p -= betaln(alpha, beta)
+
+        # Use survival value from previous period, hence the minus one
+        s = betaln(alpha, beta + num_periods - 1)
+        s -= betaln(alpha, beta)
```
CamDavidsonPilon: why put this on two lines?

lifelines/utils/ShiftedBetaGeometric.py
```diff
@@ -0,0 +1,726 @@
...
+            raise ValueError("All values of age must be equal or greater to "
+                             "one. The minimum value of "
+                             "{} was found.".format(min_age))
+
+        # Make sure alive is either zero or one!
+        alive_vals = set(alive)
+        if alive_vals != {0, 1}:
+            raise ValueError('Values for alive must be either zero or one. A '
+                             'value of '
+                             '{} was found.'.format(list(alive_vals - {0, 1})))
```
CamDavidsonPilon: elsewhere in lifelines, we don't do a check of the values, I don't think it's needed here (also users sometimes use bool/floats and not ints)

lifelines/utils/ShiftedBetaGeometric.py
```diff
@@ -0,0 +1,726 @@
...
+
+        # store the number of samples we are dealing with
+        self.n_samples = age.shape[0]
+
+        # Now that we have X we can calculate the number of parameters we need
+        n_params = X.shape[1]
+
+        # --- Optimization Starting Points
+        # Generate random starting points for optimization step.
+        initial_guesses = 0.1 * np.random.randn(restarts, 2 * n_params) - 0.01
```
CamDavidsonPilon: `- 0.01`?

lifelines/utils/ShiftedBetaGeometric.py
```diff
@@ -0,0 +1,726 @@
...
+                optimal = new_opt.fun
+                opt_params = new_opt.x
+
+            # Print current status if verbose is True.
+            if self.verbose:
+                print_string = "{step: {space}} | {elap_time:^10} | {func:13.5f} |"
+                print(print_string.format(step=step + 1,
+                                          elap_time=str(datetime.now() - start)[:-7],
+                                          func=float(optimal),
+                                          space=print_space))
```
CamDavidsonPilon: you could remove this and pass the verbose flag into `minimize`?

lifelines/utils/ShiftedBetaGeometric.py
```diff
@@ -0,0 +1,726 @@
...
+        # the scipy.optimize.minimize object in two. The first half correspond
+        # to the alpha parameter, while the second half is beta.
+        self.alpha = opt_params[:n_params]
+        self.beta = opt_params[n_params:]
+
+        # --- Regularization Penalty
+        # Compute the regularization penalty applied to the parameter vectors.
+        # Remember that there is no penalty for bias!
+        reg_penalty = self.gammaa * sum(self.alpha[1:]**2) + \
+            self.gammab * sum(self.beta[1:]**2)
```
CamDavidsonPilon: this calculation is repeated in https://github.com/CamDavidsonPilon/lifelines/pull/232/files#diff-342dd249060066abe453c23a2bc99a23R212, maybe move it into it's own function?

lifelines/utils/ShiftedBetaGeometric.py
```diff
@@ -0,0 +1,726 @@
...
+            print('{:>15}: {}'.format('wa', self.alpha))
+            print('{:>15}: {}'.format('wb', self.beta))
+            print('{:>15}: {}'.format('LogLikelihood', optimal))
+            print('{:>15}: {}'.format('Reg. Penalty', reg_penalty))
+            print()
+
+    def derl(self,
+             X,
+             age=1,
+             alive=1,
```
CamDavidsonPilon: any reason to have these as kwargs?

lifelines/fitters/SBGSurvival.py
```diff
@@ -0,0 +1,379 @@
...
+        # not the final values (since that cannot be made sense in separate).
+        suma = pd.DataFrame(data={name: (a, b) for name, a, b in
+                                  zip(self.dh.get_names(),
+                                      self.sb.alpha,
+                                      self.sb.beta)},
+                            index=['w_alpha', 'w_beta']
+                            ).T
+        return suma
+
+    def predict_params(self, df):
```
CamDavidsonPilon: should this be internal? Do the end user use this?

lifelines/fitters/SBGSurvival.py
```diff
@@ -0,0 +1,379 @@
...
+        # parameters.
+        out = pd.DataFrame(data=self.sb.churn_p_of_t(x, age=age, **kwargs))
+
+        # Give columns a decent, generic name.
+        out.columns = ['period_{}'.format(col)
+                       for col in range(1, out.shape[1] + 1)]
+
+        return out
+
+    def predict_survival(self, df, age=None, **kwargs):
```
CamDavidsonPilon: we use `predict_survival_function` elsewhere

CamDavidsonPilon:
Thanks for this contribution, @fmfn! 

As I look at this, I'm torn between what parts should live (and be exposed) in lifelines vs another package (like lifetimes). I can easily see users of lifelines being (_more_!) confused about lifelines vs lifetimes if lifelines starts to expose metrics like LTV and churn.

lifelines/utils/ShiftedBetaGeometric.py
```diff
@@ -0,0 +1,726 @@
...
+            # minimizing parameters.
+            if new_opt.fun < optimal:
+                optimal = new_opt.fun
+                opt_params = new_opt.x
+
+            # Print current status if verbose is True.
+            if self.verbose:
+                print_string = "{step: {space}} | {elap_time:^10} | {func:13.5f} |"
+                print(print_string.format(step=step + 1,
+                                          elap_time=str(datetime.now() - start)[:-7],
```
CamDavidsonPilon: I prefer to use the `time.time()` instead of datetime (lighter + no need to `str`)

lifelines/utils/ShiftedBetaGeometric.py
fmfn: No good reason, I'll put it in a single line.

lifelines/utils/ShiftedBetaGeometric.py
fmfn: My concern is that negative values will eventually raise an error when computing the beta function. Unless you know how the model work, such an error may look quite cryptic.

Perhaps I can cast alive to integer to handle both bool / float entries at once. This variable is lot more self explanatory, and erros from inputing improper values should be more self evident. You suggest simply not checking?

lifelines/utils/ShiftedBetaGeometric.py
fmfn: This version extends the original SBG with a log-linear model for predictors. So weights centred at zero are not a natural choice. The choice of `0.01` was purely driven by numerical convenience when I was experimenting with it. I am open to suggestions.

lifelines/utils/ShiftedBetaGeometric.py
fmfn: I actually think multiple restarts are best left to the user, so I'll remove this loop

lifelines/utils/ShiftedBetaGeometric.py
fmfn: 👍

lifelines/utils/ShiftedBetaGeometric.py
fmfn: Not sure I understand the question. Having them as kwargs as opposed to what?

PR #245, opened 2016-08-04
Added explicit pypi trove classifiers for supported Python versions

cfournie:
Adding this will allow libraries like [`caniusepython3`](https://pypi.python.org/pypi/caniusepython3) and [sites that rely on it](http://py3readiness.org/) to recognize that this library supports Python 3.


CamDavidsonPilon:
oh yea! 👍

PR #256, opened 2016-10-27
Fix deprecation warning

uhho:
`coerce=True` is rising deprecation warning when using pandas v0.17 and above, how about to use `errors='coerce'` instead?


CamDavidsonPilon:
looks good. CI failure issue I'll need to look into. Merging this however 🎉

PR #261, opened 2016-11-15
Update Survival Analysis intro.rst

opensorceror:
In my previous pull request, there was a typo on line 54 that causes the math markup to be displayed as plain text. My apologies. I have fixed it in this change.


CamDavidsonPilon:
👍 all good

Issue #242, opened 2016-07-28
CoxPHFitter Error


slipss:
Not sure if this is the right place for this, but I am having an issue getting the CoxPH methods to work. I am new to survival analysis, so I assume this is something wrong with my data set up. I am getting a delta contains nan's, convergence halted during the coxph.fit()? Wondering if anyone can shed some light on why this is happening?

Thanks,

ValueError                                Traceback (most recent call last)
      3 cphf1=CoxPHFitter()
----> 4 cphf1.fit(X, 'T','E')
      5 cphf1.print_summary()

/lifelines/fitters/coxph_fitter.pyc in fit(self, df, duration_col, event_col, show_progress, initial_beta, include_likelihood, strata)
    313         hazards_ = self._newton_rhaphson(df, T, E, initial_beta=initial_beta,
    314                                          show_progress=show_progress,
--> 315                                          include_likelihood=include_likelihood)
    316 
    317         self.hazards_ = pd.DataFrame(hazards_.T, columns=df.columns,

lifelines/fitters/coxph_fitter.pyc in _newton_rhaphson(self, X, T, E, initial_beta, step_size, precision, show_progress, include_likelihood)
    223             delta = solve(-h, step_size \* g.T)
    224             if np.any(np.isnan(delta)):
--> 225                 raise ValueError("delta contains nan value(s). Convergence halted.")
    226 
    227             # Save these as pending result

ValueError: delta contains nan value(s). Convergence halted.

CamDavidsonPilon:
Hey there @slipss, are you able to post the dataset you are using by chance? Or a reduced dataset?

slipss:
Not sure if this will work, but something like this, in a pandas DF

![image](https://cloud.githubusercontent.com/assets/6060449/17226195/678411d8-54c5-11e6-8cb9-930aae2a91c8.png)

CamDavidsonPilon:
Can you share that in a csv format?

slipss:
Here it is.

[Dataset_example.csv.zip](https://github.com/CamDavidsonPilon/lifelines/files/389207/Dataset_example.csv.zip)

slipss:
So, if I saved my dataframe to a csv, and then read it back into a pandas dataframe pd.read_csv() it no longer has that error.

I guess it has something to do with removing rows and columns from the dataframe, does lifelines just see those as empty perhaps?

CamDavidsonPilon:
[...]

PR #265, opened 2016-12-14
moving correct place for check of all constant columns

CamDavidsonPilon:
This will help users solve https://github.com/CamDavidsonPilon/lifelines/issues/242


lifelines/fitters/coxph_fitter.py
```diff
@@ -308,7 +310,6 @@ def fit(self, df, duration_col, event_col=None,
             df = normalize(df)
 
         E = E.astype(bool)
-        self._check_values(df)
```
CamDavidsonPilon: moving this up was the big improvement.

PR #280, opened 2017-03-22
Fixing a typo... pendantic -> pedantic

rosejn:
Thanks for the great documentation!


CamDavidsonPilon:
right, thanks the for contribution!

Issue #35, opened 2014-03-21
AalenAdditiveFitter fit is extremy slow for large matrices


aparij:
if X is large like (500k,4) , the fit method will loop over all of them ... despite the fact 99% of them the are censored.
in estimation.py the loop that I'm talking about starts on line 320 
        for i, time in enumerate(sorted_event_times):

The algebra is way beyond my understanding :) maybe I missing something but why sorted_event_times is so large ?

CamDavidsonPilon:
This is improved in v0.15

PR #286, opened 2017-04-24
solve linear system rather than explicitly computing inverse

davegolland:
[It is reported](https://github.com/CamDavidsonPilon/lifelines/issues/35) that the AalenAdditiveFitter is extremely slow for large matricies. Digging in, it seems like the solution to the ridge regression was explicitly computing the inverse, which is often [unnecessary in practice](https://www.johndcook.com/blog/2010/01/19/dont-invert-that-matrix/). This PR should reduce the runtime from O(n^3) to O(n^2) by not computing the inverse and instead solving the system of equations directly.

### DIFF
--- a/lifelines/utils/__init__.py
+++ b/lifelines/utils/__init__.py
@@ -4,7 +4,7 @@ import warnings
 from datetime import datetime
 
 import numpy as np
-from numpy.linalg import inv
+from numpy.linalg import solve
 import pandas as pd
 from pandas import to_datetime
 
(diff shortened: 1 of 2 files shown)


CamDavidsonPilon:
🎉 I love this PR! Great optimization, @davegolland 

Merging now.

davegolland:
Awesome, thanks for merging! What is the release cadence? I'm wondering when I can start relying on this patch.

CamDavidsonPilon:
Mm, a new release won't come out for a few months I suspect. You could directly use `master` or your branch in your requirement.txt files: http://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source

PR #295, opened 2017-06-04
Fix scaling in cox ph prediction

### LINKED ISSUE #294: Cox and KM give different survivals in 0.10.0
Hi,

sorry for reposting in a new issue, but I am not sure if #289 (now closed) is still being watched :)

Here are my steps:

1) prepare the data, scaling the continuous variables with sklearn's minmaxscaler. I also tried to remove my own scaling, but the results don't really change
2) pass the data through patsy and prepare the input to lifelines
3) fit a cox model
4) obtain the mean matrix (myPatsyMatrix.mean()) and use it to predict the Cox lifetime
4) perform a KM mesurement on the same population
6) compare the cox prediction and the KM measurement

With the previous version of lifelines, things were ok as long as I specified "normalize=False". Now the cox prediction and the KM measurement are significantly off.

I had a look at the changes in 0.10.0, and I am not sure I understand what is being done.

What I see in the code is:

1) you apply some transformation to the inputs (using `utils.normalize(fitData, fitDataMean, fitDataStd)`) and fit the model. this makes my own use of the minmaxscaler ineffective, but I don't see how it could create problems.
2) at prediction time, the user gives you a set of values for the features of the model. What should happen now is that you use the same transformation you used before fitting, e.g. `utils.normalize(newData, fitDataMean, fitDataStd)`, then feed the result to the calculation.

If I follow the execution flow starting e.g. from CoxPHFitter::predict_survival_function, I can't really see where you apply the transformation to the inputs X. Or am I misunderstanding the code?

Assuming I am missing something in the code and all transformations are done correctly, this means that, when predicting on a point that is exactly the mean of the fitted dataset, all partial hazards do not contribute anymore, and the only relevant term is the baseline. Doesn't this mean that the baseline survival from cox  should be fairly similar to KM's measurement?

Andrea.

CamDavidsonPilon:
cc @disimone, can you try your code with this latest branch? 

I chose to keep normalisation in the library, but it is all transparent to the user. The reason to keep it is for numeric stability.

Fixes #294

### DIFF
--- a/tests/test_estimation.py
+++ b/tests/test_estimation.py
@@ -21,7 +21,7 @@ from lifelines.estimation import CoxPHFitter, AalenAdditiveFitter, KaplanMeierFi
     NelsonAalenFitter, BreslowFlemingHarringtonFitter, ExponentialFitter, \
     WeibullFitter, BaseFitter
 from lifelines.datasets import load_larynx, load_waltons, load_kidney_transplant, load_rossi,\
-    load_lcd, load_panel_test, load_g3, load_holly_molly_polly
+    load_lcd, load_panel_test, load_g3, load_holly_molly_polly, load_regression_dataset
 from lifelines.generate_datasets import generate_hazard_rates, generate_random_lifetimes, cumulative_integral
 from lifelines.utils import concordance_index
 
@@ -92,6 +92,10 @@ def data_nus():
 def rossi():
     return load_rossi()
 
+
+@pytest.fixture
+def regression_dataset():
+    return load_regression_dataset()
 
 
 class TestBaseFitter():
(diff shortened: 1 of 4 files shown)


tests/test_estimation.py
```diff
@@ -925,12 +927,79 @@ def test_hazard_works_as_intended_with_strata_against_R_output(self, rossi):
         npt.assert_almost_equal(cp.baseline_cumulative_hazard_[(0, 0, 0, 0)].ix[[14, 35, 37, 43, 52]].values, [0.076600555, 0.169748261, 0.272088807, 0.396562717, 0.396562717], decimal=2)
         npt.assert_almost_equal(cp.baseline_cumulative_hazard_[(0, 0, 0, 1)].ix[[27, 43, 48, 52]].values, [0.095499001, 0.204196905, 0.338393113, 0.338393113], decimal=2)
 
+    def test_baseline_survival_is_the_same_indp_of_location(self, regression_dataset):
```
CamDavidsonPilon: these tests should convince you that the internal normalization is transparent to the user.

disimone:
Hi,

sorry for the late reply, I have been offline for a few days. I am happy to test this. Is the new code available in a release?

Andrea.

CamDavidsonPilon:
Hey, yes it should be as 0.10.1

disimone:
Hi,

things look a lot better now, thanks for the fix. I noticed, however, that the plot() function of the KM fitter is apparently broken.

Any attempt to use it results in a quite spectacular stack trace, but I think the culprit is in
```

KeyError                                  Traceback (most recent call last)
~/Envs/DSpy3/lib/python3.5/site-packages/matplotlib/colors.py in to_rgba(c, alpha)
    140     try:
--> 141         rgba = _colors_full_map.cache[c, alpha]
    142     except (KeyError, TypeError):  # Not in cache, or unhashable.

KeyError: (0.2980392156862745, None)
```

Plotting by hand the survival function works, so I guess the problem in really in your plotting utility.

Thanks for your help,

Andrea.

disimone:
I investigated the plotting problem. If I manually specify a color, the error disappears. this is, however, annoying, since I want to plot several lines on the same axes and let matplotlib choose the different colors automatically...

CamDavidsonPilon:
ah, strange, may be a matplotlib 2.x compatibility. Can you post your version of matplotlib here please?

disimone:
hi, I am using matplotlib 2.0.2

CamDavidsonPilon:
This kmf plot bug was fixed in 0.11.0

PR #330, opened 2017-08-27
simplify travis + drop matplotlib 1.x

CamDavidsonPilon:
this would drop matplotlib 1.x


CamDavidsonPilon:
wow so much faster builds

PR #343, opened 2017-10-15
difference also sorts columns

nonabelian:
The training set in the k fold CV has the original column order, but when picking out the test set using testing_columns, the columns get sorted by pd.index.difference. --> switched to drop.


CamDavidsonPilon:
Wow great catch. I also discovered this in a different application, but totally ignored the implications here.

CamDavidsonPilon:
Thanks @nonabelian!

Issue #351, opened 2017-11-14
Example docs: A vs. An


mattayes:
Hi folks:

There's a very minor typo in the [Examples](https://github.com/CamDavidsonPilon/lifelines/blob/master/docs/Examples.rst#plotting-multiple-figures-on-an-plot) docs that I'm happy to submit a PR for. One of the sections is called"Plotting multiple figures on an plot" when it should be "Plotting multiple figures on a plot".

mattayes:
@CamDavidsonPilon Should this issue be closed now?

PR #352, opened 2017-11-14
#351 Replace "an plot" with "a plot".

mattayes:
Small grammar fix in the Examples documentation.


CamDavidsonPilon:
Great, thanks!

PR #355, opened 2017-11-18
Adding dataset GBSG2 to lifelines datasets

klintan:
PEC library in R (and some other libraries) use GBSG2 as a sample dataset included in the packages. It would be awesome to have it in lifelines so that we can compare R-code and lifeline code so the results are similar.

I've looked through the PEC and other libraries and all of the seem to be GPL, and googled a lot to find the license for the actual data, but couldn't find any direct reference to it. If you for some reason think this is not open data (even though its available in several R packages) feel free to ignore this pull request and just leave comment. If so I will remove it myself as well and just keep it locally.

If its interesting, take a quick look on the code so everything looks ok :)


CamDavidsonPilon:
I original source looks to be 
```
W. Sauerbrei and P. Royston (1999). Building multivariable prognostic and diagnostic models: transformation of the predictors by using fractional polynomials. Journal of the Royal Statistics Society Series A, Volume 162(1), 71--94.
```

Can you add that to the doc string?

Clearly you have some R-code you would like to compare to lifelines - do you have future plans to add code for that purpose?

klintan:
I added that line to the docstring in the load_gbsg2() function :)

Yeah sort of, I'm trying to implement timeROC and integrated AUC in python. So I might be able to do a pull request for those metrics if I am successful if that is of interest ?

CamDavidsonPilon:
ah, you are correct 👍 

> Yeah sort of, I'm trying to implement timeROC and integrated AUC in python. So I might be able to do a pull request for those metrics if I am successful if that is of interest ?

This is of interest yea, look cool

PR #362, opened 2017-11-23
Exclude data storage in model

lgmoneda:
I've pickled a coxph model and it ended up with a large file size, I realized that the model stores the data inside it. This PR tries to address it, calculating training concordance and log partial hazard just after training, so it wouldn't need to store all the data anymore. I had to adapt some test cases that access the internal data too.

Is there any other reason to keep it? Is there a better way to approach this problem?

### DIFF
--- a/lifelines/fitters/coxph_fitter.py
+++ b/lifelines/fitters/coxph_fitter.py
@@ -323,9 +323,8 @@ class CoxPHFitter(BaseFitter):
         self._train_concordance = concordance_index(self.durations,
                                                     -self.predict_partial_hazard(self.data).values.ravel(),
                                                     self.event_observed)
-        self._mean_covariates = self.data.mean(0).to_frame().T
         self._train_columns = self.data.columns
-        self._train_log_partial_hazard = self.predict_log_partial_hazard(self._mean_covariates)
+        self._train_log_partial_hazard = self.predict_log_partial_hazard(self.data.mean(0).to_frame().T)
         self.data = None
         return self
 
--- a/lifelines/fitters/coxph_fitter.py
+++ b/lifelines/fitters/coxph_fitter.py
@@ -324,7 +324,8 @@ class CoxPHFitter(BaseFitter):
                                                     -self.predict_partial_hazard(self.data).values.ravel(),
                                                     self.event_observed)
         self._mean_covariates = self.data.mean(0).to_frame().T
-        self._train_columns = self.data.columns.copy()
+        self._train_columns = self.data.columns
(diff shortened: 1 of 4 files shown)


lifelines/fitters/coxph_fitter.py
```diff
@@ -299,6 +299,7 @@ def fit(self, df, duration_col, event_col=None,
         self._check_values(df, E)
         df = df.astype(float)
         self.data = df if self.strata is None else df.reset_index()
+        self._n_examples = self.data.shape[0]
```
CamDavidsonPilon: if we are to do this, I would suggest not giving CoxPHFitter the `data` attribute at all. So this would be 
```
self._n_examples = df.shape[0]

```

tests/test_estimation.py
```diff
@@ -740,7 +740,7 @@ def test_using_dataframes_vs_numpy_arrays(self, data_pred2):
         cf = CoxPHFitter()
         cf.fit(data_pred2, 't', 'E')
 
-        X = data_pred2[cf.data.columns]
+        X = data_pred2[cf._train_columns]
```
CamDavidsonPilon: I suggest not defining the `_train_columns` above, and in this test just calling `data_pred2.columns.difference(['t', 'E'])`

tests/utils/test_utils.py
```diff
@@ -458,7 +458,7 @@ def test_concordance_index_fast_is_same_as_slow():
     cp.fit(df, duration_col='week', event_col='arrest')
 
     T = cp.durations.values.ravel()
-    P = -cp.predict_partial_hazard(cp.data).values.ravel()
+    P = -cp.predict_partial_hazard(df[cp._train_columns]).values.ravel()
```
CamDavidsonPilon: same here: I would explicitly pass in the columns.

CamDavidsonPilon:
This is an interesting PR. It would break the API, so we need to proceed cautiously. I agree that saving the `data` can be large, and I've thought about removing it previously. Overall, I'm in favour of this PR, but I would suggest we merge it to `v0.13` instead of master.

CamDavidsonPilon:
For symmetry, can you use remove `self.data` from `AalensAdditiveFitter` too please?

lgmoneda:
Though the data is stored in ```AalensAdditiveFitter```, it's not used later, so I just removed the ```self.data```.

lgmoneda:
@CamDavidsonPilon is everything ok now?

CamDavidsonPilon:
lgtm, thanks @lgmoneda!

PR #370, opened 2017-12-08
V0.13

CamDavidsonPilon:
0.13 is one of lifelines biggest releases yet. 

#### 0.13.0
  - removes `is_significant` and `test_result` from `StatisticalResult`. Users can instead choose their significance level by comparing to `p_value`. The string representation of this class has changed aswell.
 - `CoxPHFitter` and `AalenAdditiveFitter` now have a `score_` property that is the concordance-index of the dataset to the fitted model.
 - `CoxPHFitter` and `AalenAdditiveFitter` no longer have the `data` property. It was an _almost_ duplicate of the training data, but was causing the model to be very large when serialized.
 - Implements a new fitter `CoxTimeVaryingFitter` available under the `lifelines` namespace. This model implements the Cox model for time-varying covariates.
 - Utils for creating time varying datasets available in `utils`.
 - less noisy check for complete separation.
 - removed `datasets` namespace from the main `lifelines` namespace
 - `CoxPHFitter` has a slightly more intelligent (barely...) way to pick a step size, so convergence should generally be faster.


docs/Examples.rst
```diff
@@ -152,8 +150,8 @@ time (months, days, ...)      observed deaths       censored
     from lifelines.utils import survival_events_from_table
 
     T, E = survival_events_from_table(df, observed_deaths_col='observed deaths', censored_col='censored')
-    print T # np.array([0,0,0,0,0,0,0,1,2,2, ...])
-    print E # np.array([1,1,1,1,1,1,1,0,1,1, ...])
+    print(T # np.array([0,0,0,0,0,0,0,1,2,2, ...]))
```
CamDavidsonPilon: wrong place for bracket

PR #379, opened 2017-12-23
Added .editorconfig

KOLANICH:


KOLANICH:
What's wrong?

CamDavidsonPilon:
Ah I should have left a note here, sorry. I think this is the wrong way to solve this problem. Rather than prescribing how an editor should behave (to influence code conventions), we should enforce code conventions with lint tests.

KOLANICH:
This is not for enforcement or prescribtion, it is for convenience only - it changes the settings of an editor to accomodate each project. For example I personally use tabs in my projects. But when I make a PR using tabs into someone's project using spaces it will likely be rejected. An I will have to either say "I'm going to maintain own fork with tabs", but it's counterproductive, or to fix the style semi-manually and send a PR again. Editorconfig setups editor for these projects, making tab key insert needed amount of spaces. Does the dirty and boring work itself.

PR #380, opened 2017-12-23
Some architecture improvements of Cox fitter

KOLANICH:
1 Splited
 * the logic determining convergence
 * the logic doing newton-raphson iteration
 * the logic doing arguments preprocessing from each other.
2 Moved some hardcoded parameters into variables.

### DIFF
--- a/lifelines/fitters/coxph_fitter.py
+++ b/lifelines/fitters/coxph_fitter.py
@@ -2,6 +2,7 @@
 from __future__ import print_function
 from __future__ import division
 
+from functools import partial
 import warnings
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
Hey @KOLANICH, thanks for the pull-request. Sorry about the delay, I've been away from my computer the last few days. 

Can you add a description to your original post with some details about what you are changing, why the change is occurring, what benefits you see from the change, etc? That would help me understand the PR more.

KOLANICH:
Just was hacking around Cox fitter trying to combine it with xgboost. Haven't managed to  implement that, but this little refactoring was a byproduct.

KOLANICH:
Are you going to merge this?

CamDavidsonPilon:
Thinking about this PR, I feel it's too early of an abstraction, especially if it doesn't open up future possibilities. There are other implementations of gradient descent in lifelines (see `CoxTimeVaryingFitter` and `WeibullFitter` we could also generalize from, but I still don't see a strong case for why.

PR #397, opened 2018-01-25
fix transform_cv_to_long_format in utils/__init__.py

OriolAndres:


OriolAndres:
in pandas 0.20.1, python 3.6.1, axis keyword is not supported as an argument to `DataFrame.rename`.

CamDavidsonPilon:
What is the behaviour in the most recent Pandas, 0.22.0?

OriolAndres:
In 0.22 both syntax formats are supported.

CamDavidsonPilon:
:+1: good with me

PR #408, opened 2018-02-20
Adding in bayesian fitter

springcoil:
Added back in Bayesian fitter. We can neglect this if you don't want it.


springcoil:
My fork was screwed up and this isn't that important. Cancelling it.

PR #409, opened 2018-02-20
Adding in bayesian fitter

springcoil:


CamDavidsonPilon:
let me know if you want a hand with this!

PR #411, opened 2018-02-21
New plots added to the Aalen Additive Fitter for better understanding…

springcoil:
… of the coefficients

I wanted to add a few things to the Aalen Additive Fitter for better understanding of the coefficients. This is for instance to debug if some of the higher order terms in your polynomial make any sense. 

What do you think Cameron?
Hattip to @dkn22 for discussions on this :)

### DIFF
--- a/lifelines/fitters/aalen_additive_fitter.py
+++ b/lifelines/fitters/aalen_additive_fitter.py
@@ -443,7 +443,8 @@ class AalenAdditiveFitter(BaseFitter):
         from matplotlib import pyplot as plt
         ax2 = kwargs.get('ax2', None) or plt.figure().add_subplot(111)
         x_bar = self.hazards_.sum(axis=0)
-        x_bar.plot(ax=ax2, kind='bar', yerr=0.1)
+        errors = self.hazards_.std(axis=0)
+        x_bar.plot(ax=ax2, kind='barh', xerr=errors)
         return ax2
 
 
--- a/lifelines/fitters/aalen_additive_fitter.py
+++ b/lifelines/fitters/aalen_additive_fitter.py
@@ -415,6 +415,38 @@ class AalenAdditiveFitter(BaseFitter):
         t = self.cumulative_hazards_.index
         return pd.DataFrame(trapz(self.predict_survival_function(X)[index].values.T, t), index=index)
 
+    @property
+    def summary(self):
+        """
+        Summary of the coefficients
+        Returns
+        -------
+        df: pd.DataFrame
+        """
+        df = self.hazards_
+        return df
+
+    def print_summary(self):
+        """
+        Print summary statistics describing the fit.
+        """
+        df = self.summary
+        print(df.to_string(float_format=lambda f: '{:4.4f}'.format(f)))
+        return
+
+    def plot_coefficients_bar_summary(self, **kwargs):
+        """
+        Plot a summary of the coefficients in Bar chart.
+        kwargs: standard Matplotlib plot arguments can be passed in
+        :return: ax2
+        """
+        from matplotlib import pyplot as plt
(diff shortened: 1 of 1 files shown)


lifelines/fitters/aalen_additive_fitter.py
```diff
@@ -415,6 +415,38 @@ def predict_expectation(self, X):
         t = self.cumulative_hazards_.index
         return pd.DataFrame(trapz(self.predict_survival_function(X)[index].values.T, t), index=index)
 
+    @property
+    def summary(self):
```
CamDavidsonPilon: usually summary has results like standard errors and p-values, I like that consistency. I would prefer to have p-values in the model before defining a summary functtion.

lifelines/fitters/aalen_additive_fitter.py
```diff
@@ -415,6 +415,38 @@ def predict_expectation(self, X):
...
+
+    def plot_coefficients_bar_summary(self, **kwargs):
+        """
+        Plot a summary of the coefficients in Bar chart.
+        kwargs: standard Matplotlib plot arguments can be passed in
+        :return: ax2
+        """
+        from matplotlib import pyplot as plt
+        ax2 = kwargs.get('ax2', None) or plt.figure().add_subplot(111)
+        x_bar = self.hazards_.sum(axis=0)
```
CamDavidsonPilon: To help the user, I'd like this to have error bars. Any way to add that?

springcoil:
I think I can add error bars let me try.

springcoil:
Any suggestions on how to add p values to the model?

springcoil:
Added error bars to this plot. Want to add p-values to the plot

lifelines/fitters/aalen_additive_fitter.py
```diff
@@ -443,7 +443,7 @@ def plot_coefficients_bar_summary(self, **kwargs):
         from matplotlib import pyplot as plt
         ax2 = kwargs.get('ax2', None) or plt.figure().add_subplot(111)
         x_bar = self.hazards_.sum(axis=0)
-        x_bar.plot(ax=ax2, kind='barh')
+        x_bar.plot(ax=ax2, kind='bar', yerr=0.1)
```
CamDavidsonPilon: is that the right error though?

springcoil: Hmm - I'm not sure. what's the right error?

springcoil: Ahh I get it :)

CamDavidsonPilon:
Looking at this PR, I'm not sure the use case of this over the normal `.plot` function. In fact, this function is just a slice of the figure `.plot` at the very last time point (when the variance is very large). Suffice to say, what does this provide over the normal `.plot`?

springcoil:
Maybe I should just close this if you think it adds no value? What do you think?

CamDavidsonPilon:
Not just yet - there was obviously an original interest of having this. I'd be open to hearing those reasons.

springcoil:
My understanding is that the plot isn't just the last time point. My understanding was that it was the sum of the coefficients. This is useful info because you'll often want to just see if these are all zero. (Unless I'm mistaken). I think p-values are too involved for this PR, so I'd prefer to be able to merge this before then. Maybe with removing some stuff and adding a TODO. Does that make sense?

CamDavidsonPilon:
> My understanding is that the plot isn't just the last time point. My understanding was that it was the sum of the coefficients.

The existing `.plot` plots the cumulative hazards. This addition plots the sum of the hazards (== to the cumulative hazard at the last point).

I wouldn't want this to ship without the proper standard errors.

springcoil:
Ok let's see if I can get proper s.e.next week.
Thanks for the clarification

springcoil:
I've not forgotten about this, I've just been busy :)

Issue #421, opened 2018-03-08
Speeding up Aalen Additive Regression


springcoil:
Hi, I've been working on a project for a few months now and one problem I have is that it can take about 4 days to run on 340k rows, with about 6 features. 

I know lifelines isn't necessarily designed for this and I've discovered that the ridge regression solve step is the biggest bottleneck - 60% of the compute time happens there. 

Are there alternative algorithms I can use like mini-batch say? Rather than the ridge regression?

CamDavidsonPilon:
Four days!? That is beyond inappropriate. I’ll look into this for the next release

springcoil:
https://stats.stackexchange.com/questions/83272/fastest-way-to-run-ridge-regression-on-large-datasets-where-np makes me think that this is a problem with my BLAS/LAPACK. Do you want me to check what they are.

springcoil:
What sort of debugging information would you need from me? How can I help you?

springcoil:
The server I was running this on - I took the spec from it. Let me know if that helps :) 


```Numexpr version:   2.6.1
NumPy version:     1.11.3
Python version:    2.7.12 | packaged by conda-forge | (default, Sep  8 2016, 14:22:31) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)]
Platform:          linux2-x86_64
AMD/Intel CPU?     True
VML available?     False
Number of threads used by default: 8 (out of 24 detected cores)

>>> np.__config__.show() 

lapack_opt_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/net/DataScience_public/jupyter_virtual_envs/python2/lib']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
blas_opt_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/net/DataScience_public/jupyter_virtual_envs/python2/lib']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
openblas_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/net/DataScience_public/jupyter_virtual_envs/python2/lib']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
openblas_lapack_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/net/DataScience_public/jupyter_virtual_envs/python2/lib']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
blas_mkl_info:
[...]

PR #423, opened 2018-03-08
BUGFIX: Removing unnecessary indentation in Aalen Additve fitter

springcoil:
I think related to https://github.com/CamDavidsonPilon/lifelines/issues/421 that there's an unnecessary indent in your fitter. Maybe I'm wrong, however consider this an attempt to fix it.

### DIFF
--- a/lifelines/fitters/aalen_additive_fitter.py
+++ b/lifelines/fitters/aalen_additive_fitter.py
@@ -183,11 +183,12 @@ class AalenAdditiveFitter(BaseFitter):
             relevant_individuals = (ids == id)
             assert relevant_individuals.sum() == 1.
 
-            # perform linear regression step.
-            try:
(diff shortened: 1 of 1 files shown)


lifelines/fitters/aalen_additive_fitter.py
```diff
@@ -183,11 +183,12 @@ def _fit_static(self, dataframe, duration_col, event_col=None,
             relevant_individuals = (ids == id)
             assert relevant_individuals.sum() == 1.
 
-            # perform linear regression step.
```
CamDavidsonPilon: the `for` loop above does nothing. I don't understand these changes

springcoil: Hmm, do you have a sketch of the algorithm for this? Now I'm confused, as it seemed that the linear regression step was causing this to hang when I ran it leading to the 4 days!!!.

springcoil:
I'm cancelling this.

Issue #418, opened 2018-03-05
Float weights in Kaplan-Meier


pzivich:
There is some unexpected behavior when fitting a weighted KM with weights that are floats. Here is a quick toy example

`import pandas as pd`
`import lifelines`
`df = pd.DataFrame()`
`df['t'] = [1,2,3,4,5]`
`df['d'] = [1,1,1,1,1]`
`df['w1'] = [2,1,1,1,1]`
`df['w2'] = [2.1,0.9,0.9,0.9,0.9]`
When fitting integer weights (w1), everything behaves as expected. However, when floats are included instead (w2) some unexpected behavior occurs. Running the following lines:
`y = lifelines.KaplanMeierFitter().fit(df['t'],weights=df['w2'])`
`print(y.survival_function_)`
Produces 
`          KM_estimate`
`timeline             `
`0.0             1.000`
`1.0             0.375`
`2.0             0.375`
`3.0             0.375`
`4.0             0.375`
`5.0             1.000`

But what is expected is:
`          KM_estimate`
`timeline             `
`0.0             1.000`
`1.0             0.632`
`2.0             0.474`
`3.0             0.316`
`4.0             0.158`
`5.0             0.000`

I haven't figured out exactly where everything goes wrong, but it looks like it is due to y.event_table converting all of the counts to int32 rather than float32. Most likely something in _preprocess_inputs is rounding everything to integers for the event_table before the KM calculation. This would explain why the survival time goes to 1 at the last time point that is less than 1.

CamDavidsonPilon:
Sorry I missed this - looks pretty important.

pzivich:
No worries. I meant to look deeper into it and find where it goes wrong, but have been pretty bogged down lately

pzivich:
Alright, I think I found where it goes wrong.  In lifelines/utils/__init__ the function survival_table_from_events() returns a pandas dataframe with all items converted to integers. It converts all the items to integers when the object is returned. Removing `.astype(int)` generates event_tables that have the correct integers when I tested the function locally. I don't know if removing this would have any downstream issues. I would guess integers was a stylistic choice?

Options are; (1) remove `.astype(int)` from a single line. Disadvantage is that event_tables won't look as clean and crisp (everything will have trailing zeroes). (2) add logic so that if weights are all integers then it is returned as integers. Otherwise, it leaves everything as a float

CamDavidsonPilon:
[...]

PR #432, opened 2018-03-26
Float weights into survival_table_from_events

pzivich:
Changing survival_table_from_events so it generates float weights and subsequent calculations properly. Edited in reference to Issue #418

### DIFF
--- a/tests/test_estimation.py
+++ b/tests/test_estimation.py
@@ -485,7 +485,7 @@ class TestKaplanMeierFitter():
             warnings.simplefilter("always")
             kmf = KaplanMeierFitter().fit(T, E, weights=np.random.random(n))
             assert True
-    def test_float_weights():
+    def test_float_weights(self):
         T = np.array([1,2,3,4,5])
         W = np.array([2.1,0.9,0.9,0.9,0.9])
         correct = np.array([1.0,0.632,0.474,0.316,0.158,0.0]).reshape((6,1))
--- a/tests/test_estimation.py
+++ b/tests/test_estimation.py
@@ -485,6 +485,13 @@ class TestKaplanMeierFitter():
             warnings.simplefilter("always")
             kmf = KaplanMeierFitter().fit(T, E, weights=np.random.random(n))
             assert True
+    def test_float_weights():
+        T = np.array([1,2,3,4,5])
+        W = np.array([2.1,0.9,0.9,0.9,0.9])
+        correct = np.array([1.0,0.632,0.474,0.316,0.158,0.0]).reshape((6,1))
+        kmf = ll.KaplanMeierFitter().fit(T,weights=W)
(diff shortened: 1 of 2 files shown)


CamDavidsonPilon:
This line is causing problems: https://github.com/CamDavidsonPilon/lifelines/pull/432/files#diff-81aff105858474f333c6fc40c9edae5eL244


Sometimes the weight, if not specified, is just 1. You could cast to an array with `np.asarray`

CamDavidsonPilon:
lgtm - I'll push it to master and may release a new bug-fix release

CamDavidsonPilon:
oh! Can you also add a test to `test_estimation.py`? Inside the `TestKaplanMeierFitter` class - somehow you found this problem existed, and I'd like to add a test to make sure it's fixed.

pzivich:
Sure thing. I have some homework to finish up first, but will work on this later today/tomorrow

pzivich:
I have not written a test within a package before. Any recommendations? Like should I input an example like my hand calculation and check whether the calculated results match what I would expect?

CamDavidsonPilon:
>  Like should I input an example like my hand calculation and check whether the calculated results match what I would expect?

yes! This is fine. I'm also okay if you would like to skip writing a test (please instead make an issue, and I can get to it when I have a chance)

pzivich:
I added a simple test to compare the results labeled as `test_float_weights()` inside the `KaplanMeierFitter` class

tests/test_estimation.py
```diff
@@ -485,6 +485,13 @@ def test_weights_can_be_floats(self):
             warnings.simplefilter("always")
             kmf = KaplanMeierFitter().fit(T, E, weights=np.random.random(n))
             assert True
+    def test_float_weights():
```
CamDavidsonPilon: this test looks great! One thing need is to include `self` in the function here (because it's a class method on `TestKaplanMeierFitter`

pzivich:
My bad. I forgot to add that in after I tested the function locally

tests/test_estimation.py
```diff
@@ -485,6 +485,13 @@ def test_weights_can_be_floats(self):
             warnings.simplefilter("always")
             kmf = KaplanMeierFitter().fit(T, E, weights=np.random.random(n))
             assert True
+    def test_float_weights(self):
+        T = np.array([1,2,3,4,5])
+        W = np.array([2.1,0.9,0.9,0.9,0.9])
+        correct = np.array([1.0,0.632,0.474,0.316,0.158,0.0]).reshape((6,1))
+        kmf = ll.KaplanMeierFitter().fit(T,weights=W)
```
CamDavidsonPilon: you don't need the `ll.` there either

CamDavidsonPilon:
tests pass! lgtm, thanks @pzivich!

PR #433, opened 2018-03-27
v0.14.1

CamDavidsonPilon:
#### 0.14.1
 - fixed bug with using weights and strata in `CoxPHFitter`
 - fixed bug in using non-integer weights in `KaplanMeierFitter`
 - Performance optimizations in `CoxPHFitter` for up to 40% faster completion of `fit`.
    - even smarter `step_size` calculations for iterative optimizations. 
    - simple code optimizations & cleanup in specific hot spots.
 - Performance optimizations in `AalenAdditiveFitter` for up to 50% faster completion of `fit` for large dataframes, and up to 10% faster for small dataframes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
  - Performance optimizations in `CoxPHFitter` for up to 40% faster completion of `fit`.
     - even smarter `step_size` calculations for iterative optimizations. 
     - simple code optimizations & cleanup in specific hot spots.
+ - Performance optimizations in `AalenAdditiveFitter` for up to 50% faster completion of `fit` for large dataframes, and up to 10% faster for small dataframes. 
 
 
 #### 0.14.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ### Changelogs
 
 #### 0.15.0
- - even smarter `step_size` calculations for iterative optimizations. 
  - fixed bug with using weights and strata in `CoxPHFitter`
+ - Performance optimizations in `CoxPHFitter` for up to 40% faster completion of `fit`.
+    - even smarter `step_size` calculations for iterative optimizations. 
+    - simple code optimizations & cleanup in specific hot spots.
 
 
 #### 0.14.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ### Changelogs
 
+#### 0.15.0
+ - even smarter `step_size` calculations for iterative optimizations. 
+ - fixed bug with using weights and strata in `CoxPHFitter`
+
+
 #### 0.14.0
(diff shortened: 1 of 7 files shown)


lifelines/fitters/aalen_additive_fitter.py
```diff
@@ -218,8 +217,9 @@ def _fit_static(self, dataframe, duration_col, event_col=None,
         self.durations = T
         self.event_observed = C
         self._compute_confidence_intervals()
+
         self.score_ = concordance_index(self.durations,
-                                        self.predict_median(dataframe).values.ravel(),
+                                        self.predict_cumulative_hazard(dataframe).values[-1,:].ravel(),
```
CamDavidsonPilon: not sure I want to keep this

lifelines/fitters/coxph_fitter.py
```diff
@@ -84,6 +88,8 @@ def fit(self, df, duration_col, event_col=None,
             self, with additional properties: hazards_
 
         """
+        #import pdb
+        #pdb.set_trace()
```
CamDavidsonPilon: rm

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ### Changelogs
 
+#### 0.15.0
```
CamDavidsonPilon: I think I'll make this 0.14.1 and remove any API changes.

lifelines/fitters/aalen_additive_fitter.py
```diff
@@ -141,14 +141,16 @@ def _fit_static(self, dataframe, duration_col, event_col=None,
...
-        pass_for_numeric_dtypes_or_raise(df)
+        self._check_values(df, T, C)
 
         ix = T.argsort()
         T, C = T.iloc[ix], C.iloc[ix]
 
         del df[duration_col]
         n, d = df.shape
         columns = df.columns
+        df = df.astype(float)
```
springcoil: I'm not sure I follow exactly what has caused the performance improvements. Is it the float? Or is it something else?

CamDavidsonPilon: It was the changes here: https://github.com/CamDavidsonPilon/lifelines/pull/433/files#diff-81aff105858474f333c6fc40c9edae5eR650

lifelines/fitters/aalen_additive_fitter.py
CamDavidsonPilon: this changes the API, hence a version upgrade.

springcoil: Hmm I think it's useful but changing the version is a big change.

lifelines/fitters/aalen_additive_fitter.py
springcoil: I just found it. So the matrix is always positive definite, that makes sense.

Issue #413, opened 2018-02-23
Aalen-Johansen Estimator


pzivich:
I looked through lifelines and I didn't see if there was implementation of an Aalen-Johansen estimator. Aalen-Johansen estimator accounts for competing risks, and is preferred over considering competing events as censored. 
You can fit this estimator by using a Kaplan-Meier to estimate the overall survival times. The survival function can then be used to generate the discrete time hazard (minus log transform S(t) and S(t-) where t- is the event time right before t, then subtract the quantities). To estimate F(t,j) you multiply S(t-) with the discrete time hazard and an indicator for j. The Aalen-Johansen estimator also cannot deal with tied event times. I wrote a function to randomly jitter the time data.

It is pretty straightforward. I wrote general code using KaplanMeierFitter() to fit the Aalen-Johansen estimator. I compared the results to SAS v9.4 on a single dataset. I would be happy to contribute it to lifelines. 

Open items: SE/CI formulas, weights, estimation methods, plotting. 

Papers detailing Aalen-Johansen or its applied use:
https://www.duo.uio.no/bitstream/handle/10852/10287/stat-res-03-97.pdf?sequence=1
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5557056/
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4325676/

PR #450, opened 2018-04-27
Create aalen_johansen_fitter.py

pzivich:
Adding a Aalen Johansen fitter as I mentioned in #413. Still needs some cleaning up. Items still needed: standard error estimator, tests, check to see how well jitter() works, ensure documentation and formatting matches rest of lifelines, write up an example

How it works is a follows: estimates an overall survival curve, calculates discrete time hazards for the event of interest (event_ind), calculates the cumulative density function (The survival function can then be used to generate the discrete time hazard (minus log transform S(t) and S(t-) where t- is the event time right before t, then subtract the quantities). To estimate F(t,j) you multiply S(t-) with the discrete time hazard and an indicator for j).

Potential addition: warn users not to calculate survival times from this (only generates the cumulative density function / risk) since the interpretation of those survival times is not straightforward. 

Some discussion and examples:
https://www.duo.uio.no/bitstream/handle/10852/10287/stat-res-03-97.pdf?sequence=1
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5557056/
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4325676/


lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import print_function
+from __future__ import division
+import numpy as np
+import pandas as pd
+
+from lifelines.fitters import KaplanMeierFitter
+from lifelines.utils import _preprocess_inputs
+
+
+class AalenJohansenFitter():
```
CamDavidsonPilon: this should subclass `UnivariateFitter`

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Class for fitting the Aalen-Johansen estimate for the cumulative incidence function in a competing 
+    risks framework.
+    
+    AalenJohansenFitter()
+    
+    """
+    def jitter(event_time,jitter_level=0.00001,seed=None):
```
CamDavidsonPilon: this is an internal function, so we should prepend the function name with with `_`

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
...
+        Returns:
+            pd.Series of jittered times
+        '''
+        if seed != None:
+            np.random.seed(seed)
+        jf = pd.DataFrame()
+        jf['t'] = event_time
+        jf['mark'] = -1 * np.random.binomial(1,0.5,size=len(jf)) #half increase, half decrease
+        jf.loc[jf['mark']==0,'mark'] = 1 
+        jf['randon_uniform'] = np.random.uniform(size=len(df)) #random amount to jitter
```
CamDavidsonPilon: do we need to add the uniform points to the dataframe? Why not:

```
jf['t2'] = jf['t'] + (jf['mark']*np.random.uniform(size=len(df))*jitter_level)
```

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
...
+            jitter_level: how much jittering to apply to data 
+        
+        Returns:
+            pd.Series of jittered times
+        '''
+        if seed != None:
+            np.random.seed(seed)
+        jf = pd.DataFrame()
+        jf['t'] = event_time
+        jf['mark'] = -1 * np.random.binomial(1,0.5,size=len(jf)) #half increase, half decrease
```
CamDavidsonPilon: I prefer using the `shape` property on a dataframe vs `len`

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
...
+        jf.loc[jf['mark']==0,'mark'] = 1 
+        jf['randon_uniform'] = np.random.uniform(size=len(df)) #random amount to jitter 
+        jf['t2'] = jf['t'] + (jf['mark']*jf['randon_uniform']*jitter_level) #jittering times
```
CamDavidsonPilon: is there a better column name than `t2`?

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
...
+        cmprisk_label = 'CDF_'+str(int(event_ind))
+        overall_events = np.where(event_observed>0,1,0) #create indicator for any event occurring
+        km = KaplanMeierFitter().fit(durations,event_observed=overall_events,timeline=timeline,entry=entry) #Fit KM for survival times
+        aj = km.event_table #extract out KM info
+        aj['Overall_S'] = km.survival_function_
+        event_spec = np.where(event_observed==event_ind,1,0) #extract observed event counts for each unique event
+        event_spec_proc = _preprocess_inputs(durations=durations,event_observed=event_spec,timeline=timeline,entry=entry,weights=weights)
+        event_spec_times = event_spec_proc[-1]['observed']
+        label_cmprisk = 'observed_'+str(int(event_ind))
+        event_spec_times.rename(label_cmprisk,inplace=True)
```
CamDavidsonPilon: lifelines avoids using `inplace`, let's do that here to

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
...
+             were born at time 0.
+          weights: weights for event
+        """
+        if np.sum(durations.duplicated()) > 0: #raise error if duplicated times
+            raise ValueError('Aalen-Johansen cannot be estimated with tied events')
+        cmprisk_label = 'CDF_'+str(int(event_ind))
+        overall_events = np.where(event_observed>0,1,0) #create indicator for any event occurring
+        km = KaplanMeierFitter().fit(durations,event_observed=overall_events,timeline=timeline,entry=entry) #Fit KM for survival times
+        aj = km.event_table #extract out KM info
+        aj['Overall_S'] = km.survival_function_
```
CamDavidsonPilon: this name is quite hard to understand - can you think of a more clear name?

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,79 @@
...
+          event_ind: integer -- indicator for event of interest. All other integers indicatin events in event observed
+             will be considered as competing events
+          timeline: return the best estimate at the values in timelines (postively increasing)
+          entry: an array, or pd.Series, of length n -- relative time when a subject entered the study. This is
+             useful for left-truncated (not left-censored) observations. If None, all members of the population
+             were born at time 0.
+          weights: weights for event
+        """
+        if np.sum(durations.duplicated()) > 0: #raise error if duplicated times
+            raise ValueError('Aalen-Johansen cannot be estimated with tied events')
```
CamDavidsonPilon: Ah, I see, one is suppose to jitter first. I think we should instead do the detection here, and _jitter if we need to (and warn the user if this occurs using the `warning` module)

pzivich:
Updated the class in regards to comments. I made _jitter() an internally called function. One item, that needs to be considered is how to allow the user to change the extent of the jittering and setting a seed. 

I have a general question: why is shape preferred over len? I haven't thought about this too much

I added references to _bounds(). I have a formula to generate the bounds, however it is not exponentiated (and therefore won't be bounded like exponential Greenwood). I either need to find one, or do the math myself one day...

I added UnivariateFitter, but I need to make sure this doesn't cause any issues and I also need to update the plots and other properties for AalenJohansen. Still need to test the function (write tests), and find a good motivating example (pull some published data)

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Class for fitting the Aalen-Johansen estimate for the cumulative incidence function in a competing 
+    risks framework.
+    
+    AalenJohansenFitter(alpha=0.95,jitter_level=0.00001,seed=None)
+    
+    Aalen-Johansen cannot deal with tied times. We can get around this by randomy jittering the event times 
+    slightly. This will be done automatically and generate a warning.
+    """
+    def fit(self,durations,event_observed,event_ind,timeline=None,entry=None,weights=None):
```
CamDavidsonPilon: from a pep8 point of view, can you add a space after each `,` - and not just on this line, but all lines of code here?

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,93 @@
...
+          entry: an array, or pd.Series, of length n -- relative time when a subject entered the study. This is
+             useful for left-truncated (not left-censored) observations. If None, all members of the population
+             were born at time 0.
+          weights: weights for event
+        """
+        #First, we check to see if tied event times, if so then need to randomly jitter the data
+        if np.sum(durations.duplicated()) > 0: #raise warning if duplicated times, then randomly jitter times
+            warnings.warn('''Tied event times were detected. The Aalen-Johansen estimator cannot handle tied event times. 
+                To resolve ties, data is randomly jittered.''',Warning)
+            durations = _jitter(event_time=durations)
```
CamDavidsonPilon: `self._jitter`?

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,93 @@
...
+    Aalen-Johansen cannot deal with tied times. We can get around this by randomy jittering the event times 
+    slightly. This will be done automatically and generate a warning.
+    """
+    def fit(self,durations,event_observed,event_ind,timeline=None,entry=None,weights=None):
+        """
+        Parameters:
+          duration: an array or pd.Series of length n -- duration of subject was observed for 
+          event_observed: an array, or pd.Series, of length n. Integer indicator of distinct events. Must be 
+             only positive integers, where 0 indicates censoring.
+          event_ind: integer -- indicator for event of interest. All other integers indicatin events in event observed
```
CamDavidsonPilon: I don't understand the role of this variable

pzivich: So ```event_ind``` was a shortened version of ```event_indicator``` is meant to indicate what the event code is in the ```event_observed``` column. For KaplanMeier or NelsonAalen ```F(t) = 1 - S(t)``` but since there are several different potential events you need to specify what event we are interested in ```F(t,j)``` where j corresponds to ```event_ind```

Imagine we have a dataset on time to AIDS among HIV infected patients where the outcomes are event_observed {AIDS = 1, Death = 2, Censored = 0}. For ```event_ind``` we would specify 1 since we are interested in time to AIDS and death is a competing risk. Does that make sense? Maybe another name for the input would be helpful

pzivich: @CamDavidsonPilon How about we name it ```event_code```. It may be more clear that you have to specify the indicator for the event of interest

pzivich:
I have update the Aalen Johansen fitter. Sorry for the long delay on this. I have the confidence intervals working. 

I need to check the plots for this, write tests, find a good data set for an example, write up a simple analysis for the website

pzivich:
It looks like v0.15.0 breaks this since ``UnivariateFitter`` lost the ``_plot_estimate`` attribute. What should this be replaced by @CamDavidsonPilon ?

Once I make that change, you can run the following code to see any example of why Aalen-Johansen should be used over Kaplan Meier when competing risks occur

```
from lifelines import AalenJohansenFitter


def competing_risk_data_generator():
    """
    Generates some random data for competing risk scenarios
    :return:
    """
    np.random.seed(1011)
    df = pd.DataFrame()
    df['T1'] = np.round(np.random.weibull(0.9, size=100) * 5, 3)
    df['T2'] = np.round(np.random.weibull(1.1, size=100) * 3, 3)
    df['C'] = np.round(np.random.weibull(1, size=100) * 10, 3)

    df['t'] = df.min(axis=1)
    df['E1'] = np.where(df['T1'] == df['t'], 1, np.nan)
    df['E1'] = np.where(df['T2'] == df['t'], 2, df['E1'])
    df['E1'] = np.where(df['C'] == df['t'], 0, df['E1'])
    df['E2'] = np.where(df['E1'] == 1, 1, 0)
    return df[['t', 'E1', 'E2']]


df = competing_risk_data_generator()
# Aalen-Johansen Fitter
aj = AalenJohansenFitter()
aj.fit(durations=df['t'], event_observed=df['E1'], event_indicator=1)
# Kaplan-Meier Fitter
km = KaplanMeierFitter()
km.fit(durations=df['t'], event_observed=df['E2'])

plt.step(aj.event_table.index, aj.cumulative_density_, where='post')
plt.step(km.event_table.index, 1-km.survival_function_, where='post')
plt.show()
```

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,170 @@
...
+        aj = aj.set_index('event_at')
+        
+        # Setting attributes
+        alpha = alpha if alpha else self.alpha
+        self._label = label
+        self.cumulative_density_ = aj[cmprisk_label] #Technically, cumulative incidence, but consistent with KaplanMeierFitter
+        self.event_table = aj[['removed','observed',self.label_cmprisk,'censored','entrance','at_risk']] #Event table
+        self.variance, self.confidence_interval_ = self._bounds(aj['lagged_overall_survival'],
+                                                                alpha=alpha, ci_labels=ci_labels)
+        self.plot = self._plot_estimate('cumulative_density_')
```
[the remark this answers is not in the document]
pzivich: That fixes it

pzivich:
No worries! It's my bad that I haven't sat down and worked on this in awhile. 

Merging into v0.15.0 sounds good to me. All the comparisons I've made with SAS check out. (SAS differs a little (+/- 0.001) but I think it is some background estimation process they have in the background. This code matches other hand-coded examples I have found and uses the same estimator). The other confidence interval method can be dealt with later (not likely important since they are approximately the same, but maybe some user would want the Aalen-based SE)

I don't have any tests written for this yet. I would have to look into how your testing is written since that's fairly new to me. I also plan on generating a simulated data set to include and updating the website documents to include this

pzivich:
Also before I forget, while I was writing this I noticed some potentially unexpected behavior in ``KaplanMeierFitter``. If you feed in an event indicator that has integers greater than 1, it reads them as events (i.e. 1's). I don't know if this behavior was intended or you want to generate a user warning if there are values outside {0, 1}. I exploited that feature to make this code a little shorter

CamDavidsonPilon:
Ah, that's expected behaviour, the event array is evaluated by its [python "truthiness"](https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-in-python-how-is-it-different-from-true-and-false), which allows user to use whatever they want (bools, ints, etc).

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,174 @@
...
+    
+    AalenJohansenFitter(alpha=0.95,jitter_level=0.00001,seed=None)
+    
+    Aalen-Johansen cannot deal with tied times. We can get around this by randomy jittering the event times 
+    slightly. This will be done automatically and generate a warning.
+    """
+    def __init__(self, jitter_level=0.0001, seed=None, alpha=0.95):
+        UnivariateFitter.__init__(self, alpha=alpha)
+        self._jitter_level = jitter_level
+        self._seed = seed  # Seed is for the jittering process
```
CamDavidsonPilon: this is never used?

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,174 @@
...
+                    -2 * sum((F_j(t) - F_j(t_i)) * S(t_i-1) * (d/n**2)
+
+        Delta method transformation:
+        SE(log(-log(F_j) = SE(F_j) / (F_j * absolute(log(F_j)))
+        
+        More information can be found at: https://support.sas.com/documentation/onlinedoc/stat/141/lifetest.pdf
+        There is also an alternative method (Aalen) but this is not currently implemented
+        """
+        # Preparing environment
+        df = self.event_table.copy()
```
CamDavidsonPilon: 👍

lifelines/fitters/aalen_johansen_fitter.py
pzivich: Ahh good catch. Didn't link that properly

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,174 @@
...
+        self._seed = seed  # Seed is for the jittering process
+
+    def fit(self,durations, event_observed, event_indicator, timeline=None, entry=None, label='AJ_estimate', 
+            alpha=None, ci_labels=None, weights=None):
+        """
+        Parameters:
+          duration: an array or pd.Series of length n -- duration of subject was observed for 
+          event_observed: an array, or pd.Series, of length n. Integer indicator of distinct events. Must be 
+             only positive integers, where 0 indicates censoring.
+          event_indicator: integer -- indicator for event of interest. All other integers are considered competing events
```
CamDavidsonPilon: how do you feel about `event_of_interest`?

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,174 @@
...
+        UnivariateFitter.__init__(self, alpha=alpha)
+        self._jitter_level = jitter_level
+        self._seed = seed  # Seed is for the jittering process
+
+    def fit(self,durations, event_observed, event_indicator, timeline=None, entry=None, label='AJ_estimate', 
+            alpha=None, ci_labels=None, weights=None):
+        """
+        Parameters:
+          duration: an array or pd.Series of length n -- duration of subject was observed for 
+          event_observed: an array, or pd.Series, of length n. Integer indicator of distinct events. Must be
```
CamDavidsonPilon: this is a neat generalization of how I usually use `event_observed` - fits nicely.

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def fit(self,durations, event_observed, event_indicator, timeline=None, entry=None, label='AJ_estimate', 
+            alpha=None, ci_labels=None, weights=None):
+        """
+        Parameters:
+          duration: an array or pd.Series of length n -- duration of subject was observed for 
+          event_observed: an array, or pd.Series, of length n. Integer indicator of distinct events. Must be 
+             only positive integers, where 0 indicates censoring.
+          event_indicator: integer -- indicator for event of interest. All other integers are considered competing events
+             Ex) event_observed contains 0,1,2 with 0:censored, 1:lung cancer, 2:death. If event_indicator=1, then deaht (2)
```
CamDavidsonPilon: typo: death

CamDavidsonPilon:
looks good. I hope this isn't too nit-picky, but can you add a space after commas, like in 
```python
event_spec = np.where(event_observed==event_indicator,1,0)

# replace with
event_spec = np.where(event_observed==event_indicator, 1, 0)
```

I'd like to see at least one test - do you want a hand setting that up?

pzivich:
``event_of_interest`` is good with me. Will update seed and fix docs typo. For testing, if you have a general idea of what you would like the test to assess, that would help me a bunch

CamDavidsonPilon:
Here's a start. The class can be placed anywhere in `tests/test_estimation.py`

After doing that, (and assuming pytest is pip-installed), you can run 
```
py.test tests/test_estimation.py -k TestAalenJohansenFitter
```
on the command line to test all the tests under the `TestAalenJohansenFitter` class. 

```python

class TestAalenJohansenFitter():
    
    @pytest.fixture  # pytest fixtures are functions that are "executed" before every test, and their output is put into the test function arguments (see tests below)
    def duration(self):
        return [1, 2, 3, 4, 5, 6]  # feel free to change these numbers
    
    @pytest.fixture
    def event_observed(self):
        return [0, 1, 1, 2, 2, 0] # feel free to change these numbers

    @pytest.fixture
    def fitter(self):
        return AalenJohansenFitter()

    def test_event_table_is_correct(self, fitter, duration, event_observed):
        fitter.fit(duration, event_observed, event_of_interest=2)

        expected_event_table = pd.DataFrame.from_records([
            # totally made up data, but needs to be corrected
            {'event_at': 1, 'removed': 1, 'observed': 0, 'censored': 1, 'entrance': 5, 'at_risk': 5, 'observed_2': 2},
            {'event_at': 2, 'removed': 1, 'observed': 1, 'censored': 0, 'entrance': 0, 'at_risk': 4, 'observed_2': 2},
            # ...
        ])
       # pandas util for checking if two dataframes are equal
        assert_frame_equal(fitter.event_table, expected_event_table, check_like=True)


    def test_variance_calcuations_against_r(self, fitter, duration, event_observed):
        pass

**Comment by pzivich** [comment]
Awesome! I will put together the tests tomorrow and push them into this PR

**Comment by CamDavidsonPilon** [comment]
Hey, FYI I changed the destination branch from `master` to `v0.15.0`

screenshot 2018-11-18 at 11 49 46

Let me know if this breaks some process on your end (I don't expect it to though)

**Comment by pzivich** [comment]
Looks like my changes to tests/test_estimation.py have some issue. I added various tests to the end of the program.

Testing assess that; _jitter works, _jitter return different times, expected event table matches, AJ is always less than or equal to KM in scenarios of competing risk, AJ is equal to KM when no competing risk, variance calculations are same as SASv9.4 using Delta method, confidence intervals from transformed SE are the same as SASv9.4. All tests passed locally for me.

File: tests/test_estimation.py [hunk 18]
```diff
@@ -1522,3 +1522,85 @@ def test_output_versus_Rs_against_standford_heart_transplant(self, ctv, heart):
...
+
+    @pytest.fixture
+    def kmfitter(self):
+        return KaplanMeierFitter()
+
+    def test_jitter(self, fitter):
+        d = pd.Series([1, 1, 1])
+        e = fitter._jitter(durations=d, event=pd.Series([1, 1, 1]), jitter_level=0.01)
+
+        npt.assert_equal(np.any(np.not_equal(d, e)), True)
```
CamDavidsonPilon: nice test

pzivich:
No problem! Happy to contribute. 

When I have some time, I will sit down and write up some documentation for the website and a simple simulated data set to demonstrate why Aalen-Johansen should be used over Kaplan-Meier in scenarios of competing risks

PR #483, opened 2018-06-12
Properly forward fill cumulative sums when making covariate timeline

rossdiener:
There is an issue where `add_covariate_to_timeline` is filling `0` values downstream of accumulated sums instead of persisting the accumulated value. For example, the expected output of cumulatively summing `covariate_one` that enters following states `covariate_one = [1, 1, 1]` should be `cumsum_covariate_one = [0, 1, 2, 3, 3, ...]` if there are additional state changes of a different covariate after these events. However, the actual output is `cumsum_covariate_one = [0, 1, 2, 3, 0, ...]`. This is easy to reproduce by running code as in the test below, which would have failed previously.

This PR fixes the issue quick and easy.

### DIFF
--- a/lifelines/utils/__init__.py
+++ b/lifelines/utils/__init__.py
@@ -1172,7 +1172,7 @@ known observation. This could case null values in the resulting dataframe."
             expanded_df['enum'] = np.arange(1, n + 1)
 
         if cumulative_sum:
-            expanded_df[cv.columns] = expanded_df[cv.columns].fillna(0)
+            expanded_df[cv.columns] = expanded_df[cv.columns].ffill().fillna(0)
 
         return expanded_df.ffill()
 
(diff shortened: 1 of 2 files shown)


lifelines/utils/__init__.py
```diff
@@ -1172,7 +1172,7 @@ def expand(df, cvs):
             expanded_df['enum'] = np.arange(1, n + 1)
 
         if cumulative_sum:
-            expanded_df[cv.columns] = expanded_df[cv.columns].fillna(0)
+            expanded_df[cv.columns] = expanded_df[cv.columns].ffill().fillna(0)
```
CamDavidsonPilon: good catch!

CamDavidsonPilon:
👍 

However, can you [point this pr](https://blog.github.com/2016-08-15-change-the-base-branch-of-a-pull-request/) at branch `v0.14.4`

PR #496, opened 2018-07-07
0.15.0

### LINKED ISSUE #188: Can't pickle univariate models properly
```
pickle.dump(kmf, open(file, 'wb' ) )

PicklingError: Can't pickle <function plot at 0x000000004D491278>: it's not found as lifelines.plotting.plot
```

### LINKED ISSUE #378: adding weights_col to docs

### LINKED ISSUE #413: Aalen-Johansen Estimator

### LINKED ISSUE #460: Bad concordance I don't know why ?
Hello, 

I have an issue  :  

here is my data  : 

image

the time is the first column and the event is the second column.

I do this code to fit the model : 

image

However i am obtaining a bad concordance 0.067  I don't know why!

image

Concordance = 0.061

### LINKED ISSUE #497: KM Late Entry
While fitting a Kaplan Meier with late entries for some simulations, I noticed an unexpected behavior. When late entries and event times occur at the same point (ex. t = 1), the late entries count towards the denominator. This is different from R and can lead to a misleading result. Example)

```
import pandas as pd 
import numpy as np 
from lifelines import KaplanMeierFitter

np.random.seed(101)

n = 10000
df = pd.DataFrame()
df['id'] = [i for i in range(n)]
df['t'] = np.ceil(np.random.weibull(1,size=n)*5)
df['cens'] = np.ceil(np.random.weibull(1,size=n)*3)
df['e'] = np.floor(np.random.weibull(1.5,size=n)*2)
df['ft'] = 10
df['obs_t'] = np.min(df[['t','cens','ft']],axis=1).astype(int)
df['d'] = (np.where(df['t']<=Ct,1,0)) * (np.where(df['t']<=df['cens'],1,0))
df['c'] = (np.where(df['cens']<=Ct,1,0)) * (np.where(df['cens']<df['t'],1,0))
df['y'] = (np.where(df['t']>df['e'],1,0)) * (np.where(df['cens']>df['e'],1,0)) * (np.where(Ct > df['e'],1,0))
df = df[['id','t','cens','ft','e','obs_t','c','d','y']].rename(columns={'e':'t_enter','obs_t':'t_out'})
dfo = df.loc[df['y']==1].copy() #"observed data"
rf = pd.DataFrame()

#Fitting KM to full data
km1 = KaplanMeierFitter()
km1.fit(df['t_out'],event_observed=df['d'])
rf['KM_true'] = km1.survival_function_

#Fitting KM to "observed" data 
km2 = KaplanMeierFitter()
km2.fit(dfo['t_out'],entry=dfo['t_enter'],event_observed=dfo['d'])
rf['KM_lifelines0.14.6'] = km2.survival_function_

#Version of KM where late entries occur after
rf['KM_lateenterafter'] = np.cumprod(1 - (km2.event_table.observed/(km2.event_table.at_risk - km2.event_table.entrance)))
print(rf)
```

Which gives

```
     KM_true  KM_lifelines0.14.6  KM_lateenterafter
0   1.000000            1.000000                NaN
1   0.818600            0.885067           0.812788
2   0.674796            0.756820           0.672143
3   0.552316            0.627204           0.548928
4   0.458322            0.520323           0.452635
5   0.375165            0.428398           0.372124
6   0.304956            0.348610           0.302718
7   0.253890            0.290921           0.252623
8   0.211575            0.242434           0.210519
9   0.174656            0.200130           0.173784
10  0.148643            0.170323           0.147901
```
My solution is a bit ugly with the NaN, but that column is the behavior I would expect. This probably went unnoticed since most data often has few tied event/entry times

### LINKED ISSUE #507: Overflow encountered in long_scalars
I've been trying to fit a CoxPHFitter on a dataset of 100.000 observations and about 15 variables (including the event- and duration columns), but I get the following warning:

```
C:\Anaconda\lib\site-packages\lifelines\utils\__init__.py:948: RuntimeWarning: overflow encountered in long_scalars
num_tied += tied
```

I've tried removing columns one by one to check for data inconsistencies, and also different transformations of variables (rounding floats, float to integer, checking for characters in numeric columns) to no avail. The samples are selected at random from a larger set of observations. 

What does work is reducing the number of observations to 78.000 (78.500 fails). After delving into the overflow issue, for example [here](https://stackoverflow.com/questions/7559595/python-runtimewarning-overflow-encountered-in-long-scalars), it seems that the num_tied is of dtype int32, and the warning is raised when the number excedes 2147483647. The fix could/should be to declare the variable like this:

`    num_tied = np.int64(0)`

Something to considere in a future update?

### LINKED ISSUE #517: kmf plotting allows inverted y-axis (to represent cumulative-ness)

### LINKED ISSUE #524: Model serialization?
First off - AWESOME library. I've been using it for a few of my projects and it's been a lifesaver. 

However, I'm looking to save models. I can't seem to find any documentation on how. My specific use case is just wanting to save the Kaplan-Meier estimator and use it to make predictions later. I can of course save off the data frame of the survival function, but I'd like to pickle (or otherwise) the model and reload in a different module as I'm performing analysis/fitting and then using the model later. Is there another way besides exporting the survival function (I'd like use the predict method...) 

Thanks!

### LINKED ISSUE #526: predict_survival_function didn't work for me
Hi, I'm not a real programmer, so sorry if this is stupid.  I tried to make a prediction based on a fit as I saw in some tutorial.  But I got the object has no attribute exp error, even though I'm fairly certain all my columns are a float or int dtype.  I even converted the boolean ones to 1s and 0s.  So then I got this to work:
```
cph = CoxPHFitter()
cph.fit(k, duration_col='conv', event_col='convb', show_progress=True, step_size = .2)
kmf = KaplanMeierFitter()
kmf.fit(k['conv'],k['convb'],label='everyone')
ax = kmf.plot()
kfake = k.drop(['conv','convb'],axis=1)

def predict_log_partial_hazard(self):
        def curry(X):
            """
            X: a (n,d) covariate numpy array or DataFrame. If a DataFrame, columns
                can be in any order. If a numpy array, columns must be in the
                same order as the training data.
            This is equivalent to R's linear.predictors.
            Returns the log of the partial hazard for the individuals, partial since the
            baseline hazard is not included. Equal to \beta (X - \bar{X})
            If X is a dataframe, the order of the columns do not matter. But
            if X is an array, then the column ordering is assumed to be the
            same as the training dataset.
            """
            if isinstance(X, pd.DataFrame):
                order = self.hazards_.columns
                X = X[order]

            index = _get_index(X)
            X = normalize(X, self._norm_mean.values, 1)
            return pd.DataFrame(np.dot(X, self.hazards_.T), index=index, dtype='float64')
        return curry
cph.predict_log_partial_hazard = predict_log_partial_hazard(cph)
cph.predict_survival_function(kfake).plot(ax=ax)
```

And I got something.  I don't know if its right, but it seems pleasant looking.

### LINKED ISSUE #529: test for cox model stratified on matched sets
- does it not fail?
 - is it performant?

### LINKED ISSUE #531: Minor typo in documentation
In the github page, as well as in the official documentation, a `plt` is missing

https://github.com/CamDavidsonPilon/lifelines/blob/master/docs/Survival%20analysis%20with%20lifelines.rst

```
ax = subplot(111)

t = np.linspace(0, 50, 51)
kmf.fit(T[dem], event_observed=E[dem], timeline=t, label="Democratic Regimes")
ax = kmf.plot(ax=ax)
print("Median survival time of democratic:", kmf.median_)
```

### LINKED ISSUE #532: Documentation not clear with time varying Cox model
At some point in the documentation, the following comment:

> From the above output, we can see that subject 1 changed state twice over the observation period, finally expiring at the end of time 10. Subject 2 was a censored case, and we lost track of them after time 2.

however, `tstop `of id 2 is 12. Should tstop be 2 instead?

### LINKED ISSUE #535: `qth_survival_time` raising ambiguous truth value error
Here's a minimal example to explain the issue:
```
from numpy.random import exponential
from lifelines import utils as survival_utils
from lifelines import KaplanMeierFitter

T_control = exponential(10, size=250)
T_experiment = exponential(20, size=200)

kmf_control = KaplanMeierFitter()
kmf_control.fit(T_control, label='control')

survival_utils.qth_survival_time(.5,kmf_control.survival_function_)
```
This raises error `The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().` because `kmf_control.survival_function_.iloc[-1]` (per [here](https://github.com/CamDavidsonPilon/lifelines/blob/master/lifelines/utils/__init__.py#L87)) returns a series, not a value.

If we replace the last line with this, it works:
`survival_utils.qth_survival_time(0.5,kmf_control.survival_function_.iloc[:,0])`

So I think we can fix this by replacing [this line](https://github.com/CamDavidsonPilon/lifelines/blob/master/lifelines/utils/__init__.py#L87) with `if survival_function.iloc[-1,0] > q:`. 

This isn't caught by the test because the test fakes a survival function with `pd.Series(somelist)`, but what we really have here is a one-column dataframe, not a series.

CamDavidsonPilon:
## v0.15.0
 
- adding `robust` params to `CoxPHFitter`'s `fit`. This enables atleast i) using non-integer weights in the model (these could be sampling weights like IPTW), and ii) mis-specified models (ex: non-proportional hazards). Under the hood it's a sandwich estimator. This does not handle ties, so if there are high number of ties, results may significantly differ from other software.
 - `standard_errors_` is now a property on fitted `CoxPHFitter` which describes the standard errors of the coefficients.
 - `variance_matrix_` is now a property on fitted `CoxPHFitter` which describes the variance matrix of the coefficients.
 - new criteria for convergence of `CoxPHFitter` and `CoxTimeVaryingFitter` called the Newton-decrement. Tests show it is as accurate (w.r.t to previous coefficients) and typically shaves off a single step, resulting in generally faster convergence. See https://www.cs.cmu.edu/~pradeepr/convexopt/Lecture_Slides/Newton_methods.pdf. Details about the Newton-decrement are added to the `show_progress` statements.
 - Minimum suppport for scipy is 1.0
 - Convergence errors in models that use Newton-Rhapson methods now throw a `ConvergenceError`, instead of a `ValueError` (the former is a subclass of the latter, however).
 - `AalenAdditiveModel` raises `ConvergenceWarning` instead of printing a warning.
 - `KaplanMeierFitter` now has a cumulative plot option. Example `kmf.plot(invert_y_axis=True)`
 - a `weights_col` option has been added to `CoxTimeVaryingFitter` that allows for time-varying weights. 
 - `WeibullFitter` has a new `show_progress` param and additional information if the convergence fails. 
 - `CoxPHFitter`, `ExponentialFitter`, `WeibullFitter` and `CoxTimeVaryFitter` method `print_summary` is updated with new fields. 
 - `WeibullFitter` has renamed the incorrect `_jacobian` to `_hessian_`. 
 - `variance_matrix_` is now a property on fitted `WeibullFitter` which describes the variance matrix of the parameters.
 - The default `WeibullFitter().timeline` has changed from integers between the min and max duration to _n_ floats between the max and min durations, where _n_ is the number of observations. 
 - Performance improvements for `CoxPHFitter` (~20% faster)
 - Performance improvements for `CoxTimeVaryingFitter` (~100% faster)
 - In Python3, Univariate models are now serialisable with `pickle`. Thanks @dwilson1988 for the contribution. For Python2, `dill` is still the preferred method.
 - `baseline_cumulative_hazard_` (and derivatives of that) on `CoxPHFitter` now correctly incorporate the `weights_col`. 
 - Fixed a bug in `KaplanMeierFitter` when late entry times lined up with death events. Thanks @pzivich
 - Adding `cluster_col` argument to `CoxPHFitter` so users can specify groups of subjects/rows that may be correlated. 
 - Shifting the "signficance codes" for p-values down an order of magnitude. (Example, p-values between 0.1 and 0.05 are not noted at all and p-values between 0.05 and 0.1 are noted with `.`, etc.). This deviates with how they are presented in other software. There is an argument to be made to remove p-values from lifelines altogether (_become the changes you want to see in the world_ lol), but I worry that people could compute the p-values by hand incorrectly, a worse outcome I think. So, this is my stance. P-values between 0.1 and 0.05 offer _very_ little information, so they are removed. There is a growing movement in statistics to shift "signficant" findings to p-values less than 0.01 anyways. 
 - New fitter for cumulative incidence of multiple risks `AalenJohansenFitter`. Thanks @pzivich! See "Methodologic Issues When Estimating Risks in Pharmacoepidemiology" for a nice overview of the model. 

closes #413 
closes #517 
closes #507
closes #460 
closes #524
closes #188 
closes #529
closes #532 
closes #531 
closes #526 
closes #497
closes #378 
closes #535


docs/Survival Regression.rst
```diff
@@ -196,21 +202,28 @@ Sometimes a covariate may not obey the proportional hazard assumption. In this c
...
 
+    ---
             coef  exp(coef)  se(coef)       z      p  lower 0.95  upper 0.95
-    fin  -0.3775     0.6856    0.1913 -1.9731 0.0485     -0.7525     -0.0024   *
-    age  -0.0573     0.9443    0.0220 -2.6081 0.0091     -0.1004     -0.0142  **
-    wexp -0.1435     0.8664    0.2127 -0.6746 0.4999     -0.5603      0.2734
-    mar  -0.4419     0.6428    0.3820 -1.1570 0.2473     -1.1907      0.3068
-    paro -0.0839     0.9196    0.1958 -0.4283 0.6684     -0.4677      0.3000
-    prio  0.0919     1.0962    0.0287  3.1985 0.0014      0.0356      0.1482  **
+    fin  -0.3788     0.6847    0.1913 -1.9799 0.0477     -0.7537     -0.0038   *
```
CamDavidsonPilon: wait why did these change?

Issue #533, opened 2018-11-13
Weibull fitter: delta nan error


hcarlens:
Hi Cam/lifelines contributors,

Thanks for this great package. 

I'm trying to use the Weibull fitter, mostly successfully, but I get an error when using this data set: 

```
from lifelines import WeibullFitter
T = [13467, 13760, 12011, 7798, 7928]
E = [0,1,0,0,0]
wf = WeibullFitter()
wf.fit(T, E)
```

This is the full error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-30-4b49a1327419> in <module>
      3 E = [0,1,0,0,0]
      4 wf = WeibullFitter()
----> 5 wf.fit(T, E)

c:\users\hcarlens\appdata\local\continuum\anaconda3\envs\lab\lib\site-packages\lifelines\fitters\weibull_fitter.py in fit(self, durations, event_observed, timeline, entry, label, alpha, ci_labels)
     99 
    100         # estimation
--> 101         self.lambda_, self.rho_ = self._newton_rhaphson(self.durations, self.event_observed)
    102         self.survival_function_ = pd.DataFrame(self.survival_function_at_times(self.timeline), columns=[self._label], index=self.timeline)
    103         self.hazard_ = pd.DataFrame(self.hazard_at_times(self.timeline), columns=[self._label], index=self.timeline)

c:\users\hcarlens\appdata\local\continuum\anaconda3\envs\lab\lib\site-packages\lifelines\fitters\weibull_fitter.py in _newton_rhaphson(self, T, E, precision)
    151             delta = solve(j, - step_size * g.T)
    152             if np.any(np.isnan(delta)):
--> 153                 raise ValueError("delta contains nan value(s). Convergence halted.")
    154 
    155             parameters += delta

ValueError: delta contains nan value(s). Convergence halted.
```

I'm using Python 3.6.6, and lifelines 0.14.6. Could you help me understand what's going wrong here? Is this a limitation in the library, or an issue with the data set? Is this something I can check for before running the fitter? 

Granted, the usefulness or meaning of the 'fit' in this specific case is debatable, but it would still be good to know why it's failing.

pzivich:
There is only one event in the data provided. The Weibull fitter can't converge to this single event. As an example, I switched the code to ``E = [0, 1, 0, 0, 1]`` and the model converged.

[...]

PR #534, opened 2018-11-14
Add Inf check for inputs

hcarlens:
As discussed in https://github.com/CamDavidsonPilon/lifelines/issues/533


hcarlens:
Looks like I missed something here - I'll fix and submit again.

CamDavidsonPilon:
Looking good so far! This is exactly what I had in mind

hcarlens:
Tests passing for all but Python 2.7 - given the number of failed tests here I'm assuming you don't expect this to succeed?

CamDavidsonPilon:
ah, I think I broke something in my pushes last night that conflict with 2.7. 

PR looks good, thanks for the addition @hcarlens!

Issue #570, opened 2018-12-12
remove deprecated `lifelines.estimation` path


CamDavidsonPilon:

usmanatron:
I'll pick this up, looks like a nice bit of refactoring!

PR #576, opened 2018-12-17
Remove lifelines.estimation

usmanatron:
Fixes issue #570 

I enjoyed that :-)
Tests pass with the same success rate as before.  Also ran prospector, no new issues reported.  Pointed at 0.16.0 branch


perf_tests/cp_perf_test.py
```diff
@@ -6,7 +6,7 @@
     import pandas as pd
     import time
 
-    from lifelines.estimation import CoxPHFitter
+    from lifelines.fitters.coxph_fitter import CoxPHFitter
```
CamDavidsonPilon: hm, does `from lifelines import CoxPHFitter` still work?

CamDavidsonPilon: that's the "new" api I want to standardize, and get rid of the "old" `.estimation` syntax

usmanatron: Oooh didn't try that.  The stuff in __init__ implies it should.  I'll have a go and update if so

usmanatron:
Thanks @CamDavidsonPilon - it's now all of the form `from lifelines import X` (except the init).

Issue #560, opened 2018-12-02
Consider refactoring high complexity methods


Kilo59:
The following methods/functions receive a mccabe [complexity score](https://en.wikipedia.org/wiki/Cyclomatic_complexity) `> 10`. This is basically a measure of loops and conditionals.

This, of course, doesn't mean that all of these methods need to be refactored, just that they are worth taking a look at.

----------------------------------------------------------------------------------------------

- [ ] mccabe: MC0001 / CoxPHFitter._newton_rhaphson is too complex (20)
- [ ] mccabe: MC0001 / CoxTimeVaryingFitter._newton_rhaphson is too complex (16)
- [ ] mccabe: MC0001 / add_covariate_to_timeline is too complex (16)
- [ ] mccabe: MC0001 / _naive_concordance_index is too complex (13)
- [ ] mccabe: MC0001 / plot_estimate is too complex (12)
- [ ] mccabe: MC0001 / AalenAdditiveFitter._fit_static is too complex (11)

-----------------------------------------------------------------------------------------------

The most important thing should be readability, maintainability and testability...not making a complexity checking tool happy 😄.

CamDavidsonPilon:
❤️❤️ this is great advice!

usmanatron:
More refactoring?!  I'm in :-)

usmanatron:
Had a stab at plot_estimate, see #581 .  @CamDavidsonPilon - I'll wait to hear your thoughts before attempting any more (to make sure you agree with my approach)

PR #581, opened 2018-12-20
Decrease the McCabe complexity of plot_estimate

usmanatron:
Fix for Issue #560 (specifically `mccabe: MC0001 / plot_estimate is too complex (12)`)

Breaking these up have been harder than I hoped :-/.  That said, Ive managed to simplify plot_estimate!
I've done this by adding a "Config" class which takes in the inputs that need pre-processing and does that separately.  Not all the inputs need this, so it's not as awful as it could've been.

Note that I haven't tested this as I don't really understand how to use lifelines.  If you're able to easily manually test that would be awesome, otherwise I can have a go (any examples would be appreciated)

### DIFF
--- a/lifelines/plotting.py
+++ b/lifelines/plotting.py
@@ -332,78 +332,102 @@ def plot_estimate(
       ax: a pyplot axis object
     """
 
-    if censor_styles is None:
-        censor_styles = {}
-
-    if (loc is not None) and (iloc is not None):
-        raise ValueError("Cannot set both loc and iloc in call to .plot().")
-
-    set_kwargs_ax(kwargs)
-    set_kwargs_color(kwargs)
-    set_kwargs_drawstyle(kwargs)
-
-    if estimate is None:
-        estimate = cls._estimate_name
-
-    if estimate == "hazard_":
-        if bandwidth is None:
-            raise ValueError("Must specify a bandwidth parameter in the call to plot_hazard.")
-        estimate_ = cls.smoothed_hazard_(bandwidth)
-        confidence_interval_ = cls.smoothed_hazard_confidence_intervals_(bandwidth, hazard_=estimate_.values[:, 0])
-    else:
-        estimate_ = getattr(cls, estimate)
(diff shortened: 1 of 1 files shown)


lifelines/plotting.py
```diff
@@ -332,78 +332,102 @@ def plot_estimate(
...
+
+        if censor_styles is None:
+            self.censor_styles = {}
+
+        if (self.loc is not None) and (self.iloc is not None):
+            raise ValueError("Cannot set both loc and iloc in call to .plot().")
+
+        set_kwargs_ax(kwargs)
+        set_kwargs_color(kwargs)
+        set_kwargs_drawstyle(kwargs)
```
CamDavidsonPilon: I know a better pattern for these, but I'll change them after this merges.

CamDavidsonPilon:
`plot_estimate` is a hairy beast, written back when I thought I was good at software engineering haha. Overall, this abstraction looks good to me - it's a nice separation of concerns. Thanks @usmanatron! Merging into 0.16.0 now.

lifelines/plotting.py
```diff
@@ -332,78 +332,102 @@ def plot_estimate(
...
-        if bandwidth is None:
-            raise ValueError("Must specify a bandwidth parameter in the call to plot_hazard.")
-        estimate_ = cls.smoothed_hazard_(bandwidth)
-        confidence_interval_ = cls.smoothed_hazard_confidence_intervals_(bandwidth, hazard_=estimate_.values[:, 0])
-    else:
-        estimate_ = getattr(cls, estimate)
-        confidence_interval_ = getattr(cls, "confidence_interval_")
-
+    config = PlotEstimateConfig(
+        cls,
```
Kilo59: I would avoid using this as a variable name especially in the context of a class method. 
https://stackoverflow.com/a/7554899/6304433

usmanatron: Good point, Ill fix this and create a new PR

PR #589, opened 2018-12-27
add pytest-icdiff

Kilo59:
Add the pytest plugin [`pytest-icdiff`](https://github.com/hjwp/pytest-icdiff) for easier to read assertion result comparisons.

image


CamDavidsonPilon:
wow cool, let me try this out a bit locally

CamDavidsonPilon:
yea I like this 👍

PR #598, opened 2019-01-07
ci_updates

Kilo59:
Because of all the checks being done in the TravisCI things are getting a little bit flakey.
Made a few changes to hopefully reduce it's flakeyness like running the linters after the build has already succeeded. 

Something that could also help is to seperate some of the responsibilities. Right knowing that the Travis build failed doesn't mean very much. It could be some minor linting/formatting error, a dependency problem or actually a major bug. 

I suggest we start using [LGTM](https://lgtm.com/) as an additional CI tool in order to reduce some of the responsibility the TravisCI has. We can then also be more strict about failed Travis builds.  
https://lgtm.com/projects/g/CamDavidsonPilon/lifelines/alerts/?mode=list
I went ahead and added badges for it to the repo, but it still requires an [admin to authorize CI access.](https://lgtm.com/projects/g/CamDavidsonPilon/lifelines/ci/)

### Changes
1. fix `pytest-cov` dependency issue
   * The version of pytest in the Travis environment was very old. Set  `pytest>=3.6`. (Latest version is `4.1`)
2. add back linting check but don't fail build because of it
3. added LGTM code quality badges
4. Fix the build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 init:
 ifeq ($(TRAVIS), true)
 		pip install -r reqs/travis-requirements.txt
+		pip list --local
 else
 		pip install -r reqs/dev-requirements.txt
 		pre-commit install
--- a/reqs/dev-requirements.txt
+++ b/reqs/dev-requirements.txt
@@ -1,10 +1,10 @@
 -r base-requirements.txt
 # installs lifelines as editiable dependancy in develop mode
 -e .
-pytest
+pytest>=3.6
 pytest-icdiff;python_version > '3.5'
 coverage>=4.4
-pytest-cov<=v2.6.0
+pytest-cov
 pypandoc
 prospector[with_pyroma]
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 init:
 ifeq ($(TRAVIS), true)
 		pip install -r reqs/travis-requirements.txt
+		pip list --local
```
Kilo59: Listing the dependencies actually be used in the build can help debug dependency issues going forward.

reqs/dev-requirements.txt
```diff
@@ -1,7 +1,7 @@
 -r base-requirements.txt
 # installs lifelines as editiable dependancy in develop mode
 -e .
-pytest
+pytest>=3.6
```
Kilo59: The Travis build environment was using a very old version of Pytest.

PR #618, opened 2019-01-28
V0.18.0

### LINKED ISSUE #465: StatisticalResult significance can be misleading
I'm not sure what the rationale behind dropping `StatisticalResult.is_significant` was, but now things are bit confusing:

* `is_significant` is still referenced in [docs/Examples.rst](https://github.com/CamDavidsonPilon/lifelines/blob/12128c0de156583b7e092eeebdbf4904a6d37468/docs/Examples.rst#L55).
* `utils.significance_code()` doesn't take the `alpha` into account, so results that are starred by `StatisticalResult.print_summary()` don't account for multiple comparisons when the Bonferroni correction is used.

### LINKED ISSUE #492: Alternatives to the logrank
Good paper here: https://arxiv.org/pdf/1807.00359.pdf

### LINKED ISSUE #518: lognormal survival curves
https://stats.stackexchange.com/questions/82611/what-is-the-formula-for-lognormal-hazard
https://core.ac.uk/download/pdf/56374045.pdf
https://github.com/MustafaOguz/Survival_Analysis/blob/master/4_Parametric_Models.ipynb

### LINKED ISSUE #572: use pytest.warns feature instead of catching warnings

### LINKED ISSUE #621: lifelines doesn't support pandas' timedelta
I am trying to generate a Kaplan Meier curve from data `df['date_of_death_or_censor']` and `df['enroll_date']` to generate a time interval to death or censor. And then `df['Dead']` a boolean series describing whether death occurred `True` or censor occurred `False`

```
from lifelines import KaplanMeierFitter
kmf = KaplanMeierFitter()
T = df['date_of_death_or_censor'] - df['enroll_date']
E = df['Dead']
kmf.fit(T, event_observed=E)
```

where `T` is dtype `timedelta64[ns]`
Generates output:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-23-882a3a6617eb> in <module>
      3 T = df['date_of_death_or_censor'] - echo['enroll_date']
      4 E = df['Dead']
----> 5 kmf.fit(T, event_observed=E)

~/.venv/jns/lib/python3.5/site-packages/lifelines/fitters/kaplan_meier_fitter.py in fit(self, durations, event_observed, timeline, entry, label, alpha, left_censorship, ci_labels, weights)
     95         # if the user is interested in left-censorship, we return the cumulative_density_, no survival_function_,
     96         estimate_name = "survival_function_" if not left_censorship else "cumulative_density_"
---> 97         v = _preprocess_inputs(durations, event_observed, timeline, entry, weights)
     98         self.durations, self.event_observed, self.timeline, self.entry, self.event_table = v
     99 

~/.venv/jns/lib/python3.5/site-packages/lifelines/utils/__init__.py in _preprocess_inputs(durations, event_observed, timeline, entry, weights)
    801         entry = np.asarray(entry).reshape((n,))
    802 
--> 803     event_table = survival_table_from_events(durations, event_observed, entry, weights=weights)
    804     if timeline is None:
    805         timeline = event_table.index.values

~/.venv/jns/lib/python3.5/site-packages/lifelines/utils/__init__.py in survival_table_from_events(death_times, event_observed, birth_times, columns, weights, collapse, intervals)
    331     death_times = np.asarray(death_times)
    332     if birth_times is None:
--> 333         birth_times = min(0, death_times.min()) * np.ones(death_times.shape[0])
    334     else:
    335         birth_times = np.asarray(birth_times)

TypeError: Cannot cast ufunc less input from dtype('<m8[ns]') to dtype('<m8') with casting rule 'same_kind'
```

CamDavidsonPilon:
- `LogNormalFitter` is a new univariate fitter you can use.
 - `WeibullFitter` now correctly returns the confidence intervals (previously returned only NaNs)
 - `WeibullFitter.print_summary()` displays p-values associated with its parameters not equal to 1.0 - previously this was (implicitly) comparing against 0, which is trivially always true (the parameters must be greater than 0)
 - `ExponentialFitter.print_summary()` displays p-values associated with its parameters not equal to 1.0 - previously this was (implicitly) comparing against 0, which is trivially always true (the parameters must be greater than 0)
 - `ExponentialFitter.plot` now displays the cumulative hazard, instead of the survival function. This is to make it easier to compare to `WeibullFitter` and `LogNormalFitter`
 - Univariate fitters' `cumulative_hazard_at_times`, `hazard_at_times`, `survival_function_at_times` return pandas Series now (use to be numpy arrays)
 - remove `alpha` keyword from all statistical functions. This was never being used. 
 - Gone are astericks and dots in `print_summary` functions that represent signficance thresholds. 
 - In models' `summary` (including `print_summary`), the `log(p)` term has changed to `-log2(p)`. This is known as the s-value. See https://lesslikely.com/statistics/s-values/
 - introduce new statistical tests between univariate datasets: `survival_difference_at_fixed_point_in_time_test`,...
 - new warning message when Cox models detects possible non-unique solutions to maximum likelihood. 
 - Generally: clean up lifelines exception handling. Ex: catch `LinAlgError: Matrix is singular.` and report back to the user advice. 

Closes #572
closes #518
closes #492
closes #465 
closes #621


lifelines/utils/__init__.py
```diff
@@ -896,6 +894,9 @@ def check_complete_separation(df, events, durations):
 
 
 def check_nans_or_infs(df_or_array):
+    if hasattr(df_or_array, "dtype") and df_or_array.dtype.kind not in "biuf":
+        raise ValueError("Values must be a subtype of number: so no strings, timedeltas, datetimes, objects, etc.")
+
```
ManuelaS: Prior to this change it was possible to use CoxPHFitter with an event information in a column of dtype object.
This was useful when calling CoxPHFitter as part of a pipeline where pre-processing steps included missing data. By using the object dtype, missing data could be represented and dropped before calling CoxPHFitter. After this change, instead the column type needs to change within the pipeline.

As a workaround, it seems that dtype float can be used throughout the pipeline.

Maybe check_nans_or_infs could be changed to accept columns of type object if all rows have acceptable content?

CamDavidsonPilon: hey @ManuelaS, happy to make changes that work with your workflow. However, can you describe the following in more detail?

> By using the object dtype, missing data could be represented and dropped before calling CoxPHFitter. After this change, instead the column type needs to change within the pipeline.

Wouldn't CoxPHFitter just blow up anyways if missing data was present?

ManuelaS: In my case, I have a sklearn pipeline with multiple steps ending with a cox model (wrapped with sklearn-lifelines). The early steps of the pipeline run on data that may have missing survival information. These rows are always dropped before CoxPHFitter is run, but with the new version dropping the rows is not sufficient - the dtype of the column also need to change.

For example: 
```
pipeline = sklearn.pipeline.Pipeline(
    steps=[
        ('consensus_cluster', py_utils.cluster_transformer.ClusterTransformer(
            py_utils.consensus_clustering.ConsensusClustering(
                n_clusters=3, base_estimator=sklearn.cluster.AgglomerativeClustering(), n_estimators=1000),
            sklearn.ensemble.RandomForestClassifier(n_estimators=1000, random_state=0))),
        ('cox', py_utils.drop_missing_estimator.DropMissingEstimator(
            sklearn_lifelines.estimators_wrappers.CoxPHFitterModel(
                duration_column='overall_survival_months', event_col='has_event')))
    ],
)
```

In this case DropMissingEstimator drops any rows with missing data in them, but is otherwise not specific to CoxPHFitterModel and does not know what the different columns in y are or what dtype they should be.

It is of course not to difficult to make this work with the new lifelines code, but I just wanted to mention it in case this was an unintended consequence of the change.

CamDavidsonPilon: I think I understand: your column has some missing data, and `DropMissingEstimator` will get rid of those, but the column dtype might still be `object` which will fail at my lifelines LOC. 

screen shot 2019-02-16 at 5 44 31 pm


Yea, this is a legit problem. I'll fix in 0.19.0 to handle this problem.

ManuelaS: Yes, that's exactly it. Thank you very much!

CamDavidsonPilon: Try the just released 0.19.0 - that should work again. Note there are some important other API changes to be aware of too: https://lifelines.readthedocs.io/en/latest/Changelog.html#api-changes

ManuelaS: Thanks!

PR #628, opened 2019-02-07
Correction for Stratified Cindex Calculation

airanmehr:
Greetings Developers,
I came across inconsistent cindex calculations in stratified-mode between lifelines and R's survival,
I looked into the issue and realized that lifelines compares pairs across strata, which it shoudn't.
I have created this solution and tested it with R and it's 100% now consistent. 
Best--
Arya

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - sudo apt-get update
 install:
   - "pip install -r dev-requirements.txt"
-  - "pip install python-coveralls pytest-cov seaborn"
+  - "pip install python-coveralls 'coverage>=4.4' pytest-cov seaborn"
 # command to run tests
 script:
   - py.test -s --cov=lifelines -vv --block=False --cov-report term-missing
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 -r requirements.txt
+coverage
 pytest
 python-coveralls
 pytest-cov


CamDavidsonPilon:
@airanmehr thanks for submitting a PR and bringing this to our attention! 

Correct me if I am wrong: R's concordance will compute concordances _within_ each strata, and then aggregate the results? 

I'm a bit surprised tbh. I'm surprised because I don't know why that's the "correct" thing to do. It seems to me that, after fitting, I can compare subjects across strata (only their baseline hazards are different, which the concordance doesn't care about.)

I'm curious more about the intuition about _why_ this is correct.

airanmehr:
@CamDavidsonPilon  Thanks for your response,
Correct, R's Survival calculates concordant and discordant pairs within each stratum and then calculates the Cindex using the aggregated numbers.

What we really wish to compare with Cindex is the risk  (which is inversely related to survival time). 
When the baseline hazard of a group are equal, we can compare their risk only by assessing their relative risk (due to PH assumption: relative risk is constant in time).
However, when comparing two groups with different baseline hazards, the relative risk is only meaningful within each stratum. Hence the right way to calculate Cindex is to count the number of concordant and discordant pairs within each stratum because the relative risk across strata is not comparable (unless their baselines are equal).

airanmehr:
It's easy to check this via an example, I had a notebook and dataset for spot checking but I can't find it right now.
Here is the code in R's package

https://github.com/cran/survival/blob/9c73189b4298b385e17e5ad381b95357c6ff24c3/R/concordance.R#L143

CamDavidsonPilon:
> When the baseline hazard of a group are equal, we can compare their risk only by assessing their relative risk (due to PH assumption: relative risk is constant in time).
However, when comparing two groups with different baseline hazards, the relative risk is only meaningful within each stratum. Hence the right way to calculate Cindex is to count the number of concordant and discordant pairs within each stratum because the relative risk across strata is not comparable (unless their baselines are equal).

Ah yes, awesome, ty. 

w.r.t to this PR, I have some other ideas on how to fix this. Do you mind if I take over? (Normally I would suggest changes for _you_ to implement, but my ideas are a bit more low-level and specific).

airanmehr:
Sure, please go ahead,
So I'll create another PR with your suggested code?

CamDavidsonPilon:
I appreciate the offer, but I will to submit a fix for the next release. I'll tag you on it however.

CamDavidsonPilon:
However, @airanmehr, one thing I would really appreciate is the notebook you mentioned above. I am having trouble getting a consistent answer in R.

airanmehr:
I will share the notebook

airanmehr:
This is the example that I generated my notebook:

http://rstudio-pubs-static.s3.amazonaws.com/5096_0880aaaf0df94f3b8533a1c024738246.html

I will redo it whenever I get a chance

CamDavidsonPilon:
Thanks for bring up this problem, @airanmehr - it's now fixed in 0.19.0 (just released)

Issue #630, opened 2019-02-10
Aalen-Johansen Variance


pzivich:
For large sample sizes, the `AalenJohansenFitter` has some trouble estimating the variances. While working on a tutorial for competing risks, I noticed some unexpected behavior for confidence intervals of the Aalen-Johansen estimator. 

This does not have to do with the formula, but rather it becomes a problem for large samples. Somewhere in the variance formulation (in `_bounds`), one element gets close to zero (or something similar). This near-zero value is causing the variance calculations to blow up. Everything works exactly as expected when n ~ 1000, but when I increase the sample size closer to n ~ 10,000 is when problems start occurring. 

Below is where the variance calculation takes place
```
            sf["part1"] = ((F_t - sf["Ft"]) ** 2) * (
                sf["observed"] / (sf["at_risk"] * (sf["at_risk"] - sf["observed"]))
            )
            sf["part2"] = (
                ((sf["lagS"]) ** 2)
                * sf[self.label_cmprisk]
                * ((sf["at_risk"] - sf[self.label_cmprisk]))
                / (sf["at_risk"] ** 3)
            )
            sf["part3"] = (F_t - sf["Ft"]) * sf["lagS"] * (sf[self.label_cmprisk] / (sf["at_risk"] ** 2))
            variance = (np.sum(sf["part1"])) + (np.sum(sf["part2"])) - 2 * (np.sum(sf["part3"]))
```
My guess is that something is going wrong with `(sf["observed"] / (sf["at_risk"] * (sf["at_risk"] - sf["observed"]))` or ` (sf[self.label_cmprisk] / (sf["at_risk"] ** 2))`, but I haven't tracked down this problem yet. The reason I target these two parts as responsible is because they will get close to zero for large sample sizes, especially the second option

When writing the original, I missed this because I tested it on sample sizes below 2,000

For replication, you can use the following data generator
```
def simulate_competing_risk(large_n):
    if large_n:
        n_idu = 1143
        n_non = 6039
    else:
        n_idu = 114
        n_non = 603

    df_idu = pd.DataFrame()
    df_idu['t_death'] = np.random.weibull(0.8, size=n_idu) * 10
    df_idu['t_aids'] = np.random.weibull(1, size=n_idu) * 10
    df_idu['idu'] = 1

    df_non = pd.DataFrame()
    df_non['t_death'] = np.random.weibull(1.1, size=n_non) * 10
[...]

Issue #629, opened 2019-02-10
Better tie handling with AalenJohansenFitter


pzivich:
`AalenJohansenFitter` currently looks for any ties in the data. We can restrict this further to reduce the calls to `_jitter`. The only ties that are really problematic for the Aalen-Johansen estimator are tied event times where the tied event times are of different types

For example, the following data would not need to have jittered event times (where 0 indicates censoring)

```
t    event
1        1
1        0
1        0
2        2
```

As it is currently written, `AalenJohansenFitter` would call `_jitter` but it is unneeded (answer would be approx equivalent). An approach that would call `_jitter` less would be to filter out censoring, then look for tied event times where the event indicator is *not* equal.

PR #632, opened 2019-02-10
Aalen johansen fixes

pzivich:
Fixes for `AalenJohansenFitter` as mentioned in #629 and #630 Pointed to master branch, since no new version branches currently available

**What this adds:**

- More specific tied event detection. The type of ties that break Aalen-Johansen are tied event times of different events. Original code detected any ties, regardless of event type. New update only looks for ties where events are not the same

- Added tests to check the new event-detection process

- Variance calculation avoids problems when large sample sizes are provided. The potentially large divisions are spaced throughout the calculation to avoid overflow. Tested on sample size of 20,000+ and had no issues

- Variance calculation takes a long time, since the calculation requires iterating over every event time. This can make processes like bootstrapping to take long then need be. In `AalenJohansenFitter.__init__` there is now the optional argument `calculate_variance` which allows the user to skip the variance calculation. Dramatically speeds up calculation time for bootstrapping

### DIFF
--- a/lifelines/fitters/aalen_johansen_fitter.py
+++ b/lifelines/fitters/aalen_johansen_fitter.py
@@ -14,14 +14,32 @@ class AalenJohansenFitter(UnivariateFitter):
     """Class for fitting the Aalen-Johansen estimate for the cumulative incidence function in a competing risks framework.
     Treating competing risks as censoring can result in over-estimated cumulative density functions. Using the Kaplan
     Meier estimator with competing risks as censored is akin to estimating the cumulative density if all competing risks
-    had been prevented. If you are interested in learning more, I (Paul Zivich) recommend the following open-access
+    had been prevented. If you are interested in learning more, we recommend the following open-access
     paper; Edwards JK, Hester LL, Gokhale M, Lesko CR. Methodologic Issues When Estimating Risks in
     Pharmacoepidemiology. Curr Epidemiol Rep. 2016;3(4):285-296.
 
-    AalenJohansenFitter(alpha=0.95, jitter_level=0.00001, seed=None)
-
     Aalen-Johansen cannot deal with tied times. We can get around this by randomy jittering the event times
     slightly. This will be done automatically and generates a warning.
+
+    AalenJohansenFitter(alpha=0.95, jitter_level=0.00001, seed=None, calculate_variance=True)
+
+    Parameters
+    ----------
+    alpha: float, option (default=0.95)
+        The alpha value associated with the confidence intervals.
+
+    jitter_level: float, option (default=0.00001)
+        If tied event times are detected, event times are randomly changed by this factor.
+
+    seed: int, option (default=None)
+        To produce replicate results with tied event times, the numpy.random.seed can be specified in the function.
+
+    calculate_variance: bool, option (default=True)
+        By default, AalenJohansenFitter calculates the variance and corresponding confidence intervals. Due to how the
+        variance is calculated, the variance must be calculated for each event time individually. This is
+        computationally intensive. For some procedures, like bootstrapping, the variance is not necessary. To reduce
+        computation time during these procedures, `calculate_variance` can be set to `False` to skip the variance
+        calculation.
     """
 
     def __init__(self, jitter_level=0.0001, seed=None, alpha=0.95, calculate_variance=True):
@@ -217,8 +235,5 @@ class AalenJohansenFitter(UnivariateFitter):
         dup_events = df.loc[df['e'] != 0, ['t', 'e']].duplicated(keep=False)
 
         # Detect duplicated times with different event types
-        ties = np.sum(dup_times & (~dup_events))
-        if ties > 0:
-            return True
-        else:
-            return False
+        ties = np.any(dup_times & (~dup_events))
+        return ties > 0
--- a/lifelines/fitters/aalen_johansen_fitter.py
+++ b/lifelines/fitters/aalen_johansen_fitter.py
@@ -69,16 +69,9 @@ class AalenJohansenFitter(UnivariateFitter):
           self, with new properties like 'cumulative_incidence_'.
         """
         # Checking for tied event times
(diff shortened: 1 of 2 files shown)


lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -205,3 +199,26 @@ def _bounds(self, lagged_survival, alpha, ci_labels):
...
+
+        # Finding duplicated event times
+        dup_times = pd.Series(df['t'].loc[df['e'] != 0]).duplicated(keep=False)
+
+        # Finding duplicated events and event times
+        dup_events = df.loc[df['e'] != 0, ['t', 'e']].duplicated(keep=False)
+
+        # Detect duplicated times with different event types
+        ties = np.sum(dup_times & (~dup_events))
+        if ties > 0:
```
CamDavidsonPilon: we can do `return ties > 0`

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -205,3 +199,26 @@ def _bounds(self, lagged_survival, alpha, ci_labels):
...
+        df['e'] = events
+
+        # Finding duplicated event times
+        dup_times = pd.Series(df['t'].loc[df['e'] != 0]).duplicated(keep=False)
+
+        # Finding duplicated events and event times
+        dup_events = df.loc[df['e'] != 0, ['t', 'e']].duplicated(keep=False)
+
+        # Detect duplicated times with different event types
+        ties = np.sum(dup_times & (~dup_events))
```
CamDavidsonPilon: np.sum -> np.any?

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -24,22 +24,23 @@ class AalenJohansenFitter(UnivariateFitter):
     slightly. This will be done automatically and generates a warning.
     """
 
-    def __init__(self, jitter_level=0.0001, seed=None, alpha=0.95):
+    def __init__(self, jitter_level=0.0001, seed=None, alpha=0.95, calculate_variance=True):
```
CamDavidsonPilon: can you add a doc string with the parameters to the above? Similar to https://github.com/CamDavidsonPilon/lifelines/blob/master/lifelines/fitters/coxph_fitter.py#L78-L98

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -184,17 +179,16 @@ def _bounds(self, lagged_survival, alpha, ci_labels):
...
-            )
-            sf["part2"] = (
-                ((sf["lagS"]) ** 2)
-                * sf[self.label_cmprisk]
-                * ((sf["at_risk"] - sf[self.label_cmprisk]))
-                / (sf["at_risk"] ** 3)
-            )
-            sf["part3"] = (F_t - sf["Ft"]) * sf["lagS"] * (sf[self.label_cmprisk] / (sf["at_risk"] ** 2))
-            variance = (np.sum(sf["part1"])) + (np.sum(sf["part2"])) - 2 * (np.sum(sf["part3"]))
+            variance = np.sum((F_t - sf["Ft"]) ** 2 *
```
CamDavidsonPilon: did `black` do this???

tests/test_estimation.py
```diff
@@ -3293,10 +3293,25 @@ def test_jitter(self, fitter):
...
+        # Based on new setup of ties, this counts as a valid tie
         d = [1, 2, 2, 4, 5, 6]
         fitter.fit(durations=d, event_observed=[0, 1, 2, 1, 2, 0], event_of_interest=2)
         npt.assert_equal(np.any(np.not_equal([0] + d, fitter.event_table.index)), True)
 
+    def test_updated_input_ties(self, fitter):
+        # Based on the new setup of ties, should not detect any ties as existing
+        d = [1, 2, 2, 4, 5, 6]
+        fitter.fit(durations=d, event_observed=[0, 1, 1, 1, 2, 0], event_of_interest=1)
+        print(fitter.event_table.index)
```
CamDavidsonPilon: rm, and the below print too

lifelines/fitters/aalen_johansen_fitter.py
pzivich: If you are referring to `variance` being collapsed into a single calculation (rather than in three pieces) that was my change.

The spacing may either be `black` or my IDE (PyCharm)

CamDavidsonPilon: it's a really difficult to read piece of code. Can you break it back up a bit by using intermediate values? For example: https://github.com/CamDavidsonPilon/lifetimes/blob/master/lifetimes/fitters/beta_geo_fitter.py#L156-L166

lifelines/fitters/aalen_johansen_fitter.py
```diff
@@ -14,32 +14,51 @@ class AalenJohansenFitter(UnivariateFitter):
     """Class for fitting the Aalen-Johansen estimate for the cumulative incidence function in a competing risks framework.
     Treating competing risks as censoring can result in over-estimated cumulative density functions. Using the Kaplan
     Meier estimator with competing risks as censored is akin to estimating the cumulative density if all competing risks
-    had been prevented. If you are interested in learning more, I (Paul Zivich) recommend the following open-access
```
CamDavidsonPilon: lol ok 👍

CamDavidsonPilon: put your phone number in

CamDavidsonPilon: lol jk

pzivich: Didn't want to get you in trouble with my esoteric paper recommendations... I will put my home address in as well

CamDavidsonPilon:
lgtm, thanks @pzivich! Don't worry about the single test failure (it's a lint thing). I'll get this into the next release today or tomorrow 👍

PR #678, opened 2019-03-25
Add variance formula for Kaplan Meier estimator

syadlowsky:


CamDavidsonPilon:
👋 hello @syadlowsky! 

Thanks for the contribution - can you expand on why you would like to see this added?

syadlowsky:
I'm working on a project where, to do inference, we need to be able to estimate the variance of the Kaplan-Meier estimator. It looks like right now, you only have the confidence intervals, but exposing the variance estimator itself might be useful for others' downstream applications, as well.

syadlowsky:
Just wanted to bring this up again.

syadlowsky:
Somehow, I didn't see the reply. Sorry about that. The issue with the exponential form of the variance is that it gives a variance for log(S(t)), whereas I need (and one might expect a variance function to give) the variance for S(t).

One alternative is to add both, adding the exponential form as variance_log_survival_. At least this gives something clear for people to think about when choosing which they want, and would possibly allow some code reuse with the confidence interval estimates. Plus it would give something to point to if someone asks why variance_ doesn't line up with confidence intervals.

Does this seem like reasonable recourse? If so, I can fix this and try to deal with the merge conflict that has come up in the meantime.

PR #682, opened 2019-03-27
Piecewise regression

CamDavidsonPilon:


snoodleboot:
This bitt me. 

    def __init__(self, breakpoints, *args, **kwargs):
        breakpoints = np.sort(breakpoints)
        if breakpoints and not (breakpoints[-1] < np.inf):
            raise ValueError("Do not add inf to the breakpoints.")

in your cotr the condition "breakpoints and (breakpoints[-1] < np.inf) should be "not (breakpoints and (breakpoints[-1] < np.inf))" since you are wanting to make sure that if breakpoints, then the endpoint cannot be inf. This is an issue if breakpoints is None. Better to have chained conditional statements. I will write a change and PR it.

PR #708, opened 2019-04-12
removing axes from errorbar_kwargs if provided

### LINKED ISSUE #707: Cox time varying fitter plot method raises TypeError if `ax` is provided.
Minimal example (adapted for the docs):

```python
import matplotlib.pyplot as plt
from lifelines import CoxTimeVaryingFitter

ax = plt.figure().add_subplot(111)
ctv = CoxTimeVaryingFitter()
ctv.fit(df, id_col="id", event_col="event", start_col="start", stop_col="stop", show_progress=True)
ctv.plot(ax=ax)
```
Leads to:
```python
TypeError: inner() got multiple values for argument 'ax'
```

**Solution  (suggestion)**:

Use `pop` instead of `get` on `errorbar_kwargs` within `CoxTimeVaryingFitter.plot`.

j-i-l:
Should fix #707

### DIFF
--- a/lifelines/fitters/cox_time_varying_fitter.py
+++ b/lifelines/fitters/cox_time_varying_fitter.py
@@ -723,7 +723,7 @@ See https://stats.stackexchange.com/questions/11109/how-to-deal-with-perfect-sep
         """
         from matplotlib import pyplot as plt
 
-        ax = errorbar_kwargs.get("ax", None) or plt.figure().add_subplot(111)
+        ax = errorbar_kwargs.pop("ax", None) or plt.figure().add_subplot(111)
 
         errorbar_kwargs.setdefault("c", "k")
         errorbar_kwargs.setdefault("fmt", "s")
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
:+1: yup thanks for this fix

I wonder if CoxPHFitter has the same problem?

j-i-l:
no worries.
I've changed other minor things that might cause issues when integrating plots in existing figures, like the `plt.yticks` calls instead of using the current axes with `ax.set_yticks` + `ax.set_yticklabels`, etc.

If I find the time, I'll have a quick look at the other plot functions.

j-i-l:
> +1 yup thanks for this fix
> 
> I wonder if CoxPHFitter has the same problem?

No, CoxPHFitter is fine. After your `plotting.set_kwargs_ax` call, `errorbar_kwargs` has an axes that you then `pop`. 
Might be a good idea to use the same approach for both though.

PR #719, opened 2019-05-15
Changed the way printing is done through iterations by flushing stdout

sachinruk:
Before with an iteration, all the metrics were taking a new line. Now the numbers are printed in one line and can be seen as they change. Easier to look at especially when looking at Jupyter notebooks.

### DIFF
--- a/lifelines/fitters/aalen_additive_fitter.py
+++ b/lifelines/fitters/aalen_additive_fitter.py
@@ -248,7 +248,8 @@ class AalenAdditiveFitter(BaseFitter):
             X[exits, :] = 0
 
             if show_progress and i % int((n_deaths / 10)) == 0:
-                print("Iteration %d/%d, seconds_since_start = %.2f" % (i + 1, n_deaths, time.time() - start))
+                print("\rIteration %d/%d, seconds_since_start = %.2f" % (i + 1, n_deaths, time.time() - start),
+                      end="")
 
             last_iteration = i + 1
             # terminate early when there are less than (3 * d) subjects left, where d does not include the intercept.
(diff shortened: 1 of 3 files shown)


lifelines/fitters/aalen_additive_fitter.py
```diff
@@ -248,7 +248,8 @@ def _fit_model_to_data_batch(self, X, T, E, weights, show_progress):
             X[exits, :] = 0
 
             if show_progress and i % int((n_deaths / 10)) == 0:
-                print("Iteration %d/%d, seconds_since_start = %.2f" % (i + 1, n_deaths, time.time() - start))
+                print("\rIteration %d/%d, seconds_since_start = %.2f" % (i + 1, n_deaths, time.time() - start),
```
CamDavidsonPilon: 👍

lifelines/fitters/weibull_aft_fitter.py
```diff
@@ -169,3 +169,54 @@ def predict_cumulative_hazard(self, X, times=None, ancillary_X=None):
         lambda_, rho_ = self._prep_inputs_for_prediction_and_return_scores(X, ancillary_X)
 
         return pd.DataFrame(np.outer(times, 1 / lambda_) ** rho_, columns=_get_index(X), index=times)
+
+    def predict_cumulative_hazard_at_single_time(self, X, times, ancillary_X=None):
```
CamDavidsonPilon: I'd like to not include these in this PR (I assume it wasn't intended to be included?)

sachinruk: Whoops, yeah I was going to submit that as my next PR (after our conversation on SO), would you want a PR for that, or do you want to implement that functionality differently.

CamDavidsonPilon:
lgtm, thanks!

lifelines/fitters/weibull_aft_fitter.py
CamDavidsonPilon: I would like to think further about how to include this functionality, so let's leave it out for now.

PR #726, opened 2019-05-27
Fixed scipy version to <=1.2.1

Sikerdebaard:
Fixed scipy version to <= 1.2.1 until autograd HIPS/autograd#490 is released. After that we could fix autograd to >= whatever version they release with that specific PR and return scipy to >=1.0.

In scipy >= 1.3.0 some functions that autograd uses are moved to a different namespace. Due to this change lifelines currently will not work when installed trough a package manager like pypi.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     ],
     install_requires=[
         "numpy>=1.6.0",
-        "scipy>=1.0",
+        "scipy<=1.2.1",
         "pandas>=0.23.0",
         "matplotlib>=3.0",
         "bottleneck>=1.0",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -37,7 +37,7 @@ def filepath(fname):
     ],
     install_requires=[
         "numpy>=1.6.0",
-        "scipy>=1.0",
+        "scipy<=1.2.1",
```
CamDavidsonPilon: can pip handle lower _and_ upper bounds? I'd still like the requirement that scipy >= 1.0.

CamDavidsonPilon:
Thanks for PRing this!

The current changes won't change the setup on TravisCI - we'll also need to change the scipy in `base-requirements.txt`

Sikerdebaard:
Yup, pip can have min max bounds. I've added those and also modified the base-requirements.txt.

PR #729, opened 2019-05-31
Fixing reference in warning text for "unexpected datetype"

ktr:
The warning text refers to `lifelines.utils.datetime_to_durations` but should refer to `lifelines.utils.datetimes_to_durations` (notice "s" after datetime).


CamDavidsonPilon:
Great, ty!

PR #736, opened 2019-06-04
Added an adapter allowing to plug lifelines models to the stuff expecting sklearn models

### LINKED ISSUE #553: sklearn.model_selection compatibility
sklearn.model_selection has bunch of useful functions, like `cross_val_predict` useful for visualization of predicted times against actual ones.

KOLANICH:
... s.a. crossvalidation and `cross_val_predict`.
Fixes #553

### DIFF
--- a/lifelines/utils/sklearnAdapter.py
+++ b/lifelines/utils/sklearnAdapter.py
@@ -1,0 +1,81 @@
+from sklearn.base import BaseEstimator, RegressorMixin, MetaEstimatorMixin, MultiOutputMixin
+from sklearn.model_selection import cross_validate
+from sklearn.metrics import fbeta_score, make_scorer
+from . import concordance_index
+import pandas
+
+import inspect
+
+
+def filterKwArgs(f, kwargs):
+    s = inspect.getfullargspec(f)
+    if s.varkw:
+        return kwargs
+    else:
+        allArgs = set()
+        if s.args:
+            allArgs |= set(s.args)
+        if s.kwonlyargs:
+            allArgs |= set(s.kwonlyargs)
+        fedArgs = set(kwargs)
+        redundantArgs = fedArgs - allArgs
+        presentArgs = fedArgs & allArgs
+        #if redundantArgs:
+        #    warnings.warn("Following args are redundant for "+str(f)+str(inspect.signature(f))+": "+repr(redundantArgs))
+        res = {k: kwargs[k] for k in presentArgs}
+        return res
+
+
+class LifelinesSKLearnAdapter(BaseEstimator, MetaEstimatorMixin, RegressorMixin):
+    def __init__(self, fitter, params, yArgName="duration_col", eventArgName="event_col", remap=None):
+        self.fitter = fitter
+        assert yArgName in params
+        self.params = params
+        self.yArgName = yArgName
+        self.eventArgName = eventArgName
+        if remap is None:
+            remap = {}
+        self.remap = remap
+
+    @property
+    def yColumn(self):
+        return self.params[self.yArgName]
+
+    @property
+    def eventColumn(self):
+        return self.params[self.eventArgName] if self.eventArgName else None
+
+    def fit(self, X, y=None, sample_weight=None):
+        """y has the default value to allow a keyword one because sklearn pass the stuff as to a keyword one"""
+        if y is not None:
+            X.insert(len(X.columns), self.yColumn, y, allow_duplicates=False)
+
+        params = type(self.params)(self.params)
+
+        for nm, newNm in self.remap.items():
+            params[newNm] = params[nm]
+            del params[nm]
+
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
Neat!!! This is such a simple idea! I'll review later tonight

lifelines/utils/sklearnAdapter.py
```diff
@@ -0,0 +1,84 @@
+import inspect
+import pandas
+from sklearn.base import BaseEstimator, RegressorMixin, MetaEstimatorMixin, MultiOutputMixin
```
CamDavidsonPilon: scikit learn isn't a dependency for the project, and I don't want to natively make it one. Can we do a try-catch import here, with an error message saying to install scikit-learn?

lifelines/utils/sklearnAdapter.py
```diff
@@ -0,0 +1,84 @@
+import inspect
```
CamDavidsonPilon: my file name convention has been lower-case underscores. Can you duplicate that here? So `sklearn_adapter.py`

lifelines/utils/sklearnAdapter.py
```diff
@@ -0,0 +1,84 @@
...
+        presentArgs = fedArgs & allArgs
+        #if redundantArgs:
+        #    warnings.warn("Following args are redundant for "+str(f)+str(inspect.signature(f))+": "+repr(redundantArgs))
+        res = {k: kwargs[k] for k in presentArgs}
+        return res
+
+
+class LifelinesSKLearnAdapter(BaseEstimator, MetaEstimatorMixin, RegressorMixin):
+    __slots__ = ("fitter", "params", "yArgName", "eventArgName", "remap", "concordanceRemap")
+    def __init__(self, fitter, params, yArgName="duration_col", eventArgName="event_col", remap=None, concordanceRemap=None):
```
CamDavidsonPilon: can you add a doc string for this class? 

Also, can we use snakecase for the parameter names?

lifelines/utils/sklearnAdapter.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.params[self.yArgName]
+
+    @property
+    def eventColumn(self):
+        return self.params[self.eventArgName] if self.eventArgName else None
+
+    def fit(self, X, y=None, sample_weight=None):
+        """y has the default value to allow a keyword one because sklearn pass the stuff as to a keyword one"""
+        if y is not None:
+            X.insert(len(X.columns), self.yColumn, y, allow_duplicates=False)
```
CamDavidsonPilon: is `insert` inplace? If not, we need `X = X.insert...` here

CamDavidsonPilon:
It would be great if you could add tests, else I can handle that. If not, can you add a simple script / jupyter notebook of using this in the `experiments/` or `examples/` folder?

lifelines/utils/sklearnAdapter.py
KOLANICH: This module is not imported automatically when importing `lifelines.utils` or `from lifelines.utils import *` because it is not in `__all__`. And it is useful only for the ones already using `sklearn`. So if a person imports this module he is expected to have `sklearn`. Otherwise he just doesn't import it and `sklearn` is not needed;

lifelines/utils/sklearnAdapter.py
KOLANICH: feel free to do all the fixes you feel needed.

lifelines/utils/sklearnAdapter.py
KOLANICH: The same. Feel free to rename them the way you like. It is your lib, not mine.

KOLANICH:
>If not, can you add a simple script / jupyter notebook of using this in the experiments/ or examples/ folder?

Certainly I can. But you need to determine the final variables names first. About the test: I have tested by doing Cox regression on `rossi` dataset and comparing the results of `k_fold_cross_validate` to `sklearn.model_selection.cross_validate`. I guess it is not a very good test - it depends on other parts of this lib. If you need something better, you would have to write it yourself.

lifelines/utils/sklearnAdapter.py
KOLANICH: If it wasn't inplace, there would have been no sense to have this method. It returns `None`:

https://github.com/pandas-dev/pandas/blob/e7ad88441b2586bc7f725ecd9b354b35a3ec8c99/pandas/core/frame.py#L3437L3456

CamDavidsonPilon:
Okay, well this is a good start. It's easiest for me to add to it once it is in master, so I'll merge and make the desired changes. I'll add tests as well.

CamDavidsonPilon:
If you do have any script / code of this in use, that would be helpful to see as well.

PR #743, opened 2019-06-11
Replaced setup.py to setup.cfg for storage of metadata.

KOLANICH:
Version is now fetched from git tag.
Lifted restriction on scipy version because `worksforme` (tests pass) at the latest one.


CamDavidsonPilon:
> Version is now fetched from git tag.

where does that part live?

KOLANICH:
In a git tag. You have tags like `v0.1.something` . `setuptools_scm` selects the closest such tag and parses version from it.

KOLANICH:
@CamDavidsonPilon, ignoring PRs for years and then silently closing them is just impolite.

PR #744, opened 2019-06-11
Some improvements in the adapter

### LINKED ISSUE #553: sklearn.model_selection compatibility

KOLANICH:
Fixes #553

Though these improvements in fact may interfere with the subclass use case (I have not mind it a lot and I still think that dynamically generated classes for each fitter class are undesireable, and so is messing with globals)


KOLANICH:
@CamDavidsonPilon, ignoring PRs for years and then silently closing them is just impolite.

PR #757, opened 2019-06-30
V0.22.0

### LINKED ISSUE #695: Gompertz models
https://spia.uga.edu/faculty_pages/rbakker/pols8501/OxfordTwoNotes.pdf

### LINKED ISSUE #724: More than two parameters in an AFT model
Hi,
I am trying to use a "Maximal Weibull" distribution for an AFT model. This distribution has three parameters (`lambda`, `beta` and `delta`). I have successfully implemented the corresponding parametric univariate fitter with these three parameters. However, I am puzzled about how to proceed with the AFT model, as all the implemented models have two parameters only (a primary and an ancillary one). I suppose that this should be in theory possible, right? Would it require major changes to the package code to do it?

PS: I know you didn't plan on people implementing their own AFT models, but well, your package is so well implemented that it invites you to do it ;)

### LINKED ISSUE #733: start to deprecate `cph.hazards_` in favour of `params_`
`hazards_` is misleading, I feel. It's too close to _hazard ratios_, and the attached values are _not_ hazard ratios. `params_` is used by the AFT models, so it makes sense to align there.

### LINKED ISSUE #734: In `.confidence_intervals_`, put confidence level in column names.

### LINKED ISSUE #735: in `.summary`, include `exp(param)` confidence intervals
And use this to clean up the `hazard_ratios` in `plot` code.

### LINKED ISSUE #751: Generalizing Parametric Regression models
### Design goals:

1) Easy to add custom regression models, similar to how one can create custom univariate models.

2) Arbitrary number of parameters to fit to. Example: an exponential AFT model has a single parameter, whereas Weibull AFT model has two, and a custom one could have N. 

3) Fixing parameters. I should be able to fix n < N parameters to set parameters.

4) custom penalizers (this enables point 5 below) for users to allow some interesting parameter/information sharing. 

5) Stretch goal: make PiecewiseExponentialRegressionFitter less of a snowflake. 

Point 2. and 3. allow the Exponential AFT fitter to be a special case of the the Weibull AFT fitter. 

### Prior workarounds or issues

See also #720 and [this answer](https://stackoverflow.com/questions/56787015/is-it-possible-to-run-a-cox-proportional-hazards-model-with-an-exponential-distr/56794328#56794328)

### Some thoughts:

- I think I need to expel the idea of "primary" and "ancillary" for the general case, and make `Known` models reimplement those
- `PiecewiseExponentialRegressionFitter` is currently a snowflake. One of the reasons was because it has an arbitrary number of parameters and a customer penalizer function. I might be able to fold that into this class. 

### Custom AFT API

```python

class ShiftedWeibull(ParametericRegressionFitter):

    _fitted_parameter_names = ['lambda_', 'rho_', 'delta_']
    # no point in _bounds: assume they range over all floats, the user can change to positive only 
    # using the cumulative hazard and exp()

    def _cumulative_hazard(self, params, T, Xs):
        lambda_ = np.exp(np.dot(Xs["lambda_"], params["lambda_"])) # > 0
        rho_    = np.exp(np.dot(Xs["rho"],     params["rho_"]))    # > 0
        delta_  =        np.dot(Xs["delta_"],  params["delta_"])

        return ((T - delta_) / lambda_) ** rho_


    def _add_penalty(self, neg_ll, params):
        """ 
        only penalize the delta parameter, for whatever reason
        """
        penalty = params['delta_'] ** 2
        return neg_ll + self.penalizer * penalty



swf = ShiftedWeibull(penalizer=1.0)

covariates = {
  'lambda_': [variable names], # need a shortcut for all columns?
  'rho_':    [variable names], 
  'delta_':  [variable names] # (or fixed constant),
}

swf.fit_right_censoring(df, "T", "E", regressors=covariates) # TODO: name 
```

### LINKED ISSUE #754: Save KaplanMeierFitter model to disk
Hi, 

Is there any way to persis the KMF model on disk like sklearn functionality?

### LINKED ISSUE #760: for parametric univariate models, make conditional_time_to_event_ "smarter"
Ex: If I have a weibull parametric model, I should be able to exactly calculate the conditional time to event without relying on the `survival_function_` or `timeline`

CamDavidsonPilon:
##### New features
 - Ability to create custom parametric regression models by specifying the cumulative hazard. This enables new and extensions of AFT models.
 - `percentile(p)` method added to univariate models that solves the equation `p = S(t)` for `t`
 - for parametric univariate models, the `conditional_time_to_event_` is now exact instead of an approximation.

##### API changes
 - In Cox models, the attribute `hazards_` has been renamed to `params_`. This aligns better with the other regression models, and is more clear (what is a hazard anyways?)
 - In Cox models, a new `hazard_ratios_` attribute is available which is the exponentiation of `params_`.
 - In regression models, the column names in `confidence_intervals_` has changed to include the alpha value.
 - In regression models, some column names in `.summary` and `.print_summary` has changed to include the alpha value.
 - In regression models, some column names in `.summary` and `.print_summary` includes confidence intervals for the exponential of the value.
 - Significant changes to internal AFT code.
 - A change to how `fit_intercept` works in AFT models. Previously one could set `fit_intercept` to False and not have to set `ancillary_df` - now one must specify a DataFrame.

##### Bug fixes
 - for parametric univariate models, the `conditional_time_to_event_` is now exact instead of an approximation.

closes #760
closes #754 
closes #751
closes #735
closes #734
closes #733
closes #724
closes #695


CamDavidsonPilon:
GH seems to be behind. This is green: https://travis-ci.org/CamDavidsonPilon/lifelines/builds/553900497

PR #758, opened 2019-07-02
Fix typo in smoothing hazard curve docs

dmedvinsky:
There is no such method `hazard_confidence_intervals_`,
but there is a `smoothed_hazard_confidence_intervals_`.


CamDavidsonPilon:
👌 nice catch, thanks!

PR #766, opened 2019-07-06
Gamma experiments

CamDavidsonPilon:


experiments/gamma_experiments.py
```diff
@@ -0,0 +1,47 @@
...
+
+df = load_waltons()
+
+gg = GG()
+gg.fit(df["T"], df["E"])
+
+gg.print_summary()
+
+
+class LogGammaFitter(ParametericUnivariateFitter):
```
CamDavidsonPilon: @cristobaltapia here's an implementation of your model using a slightly more stable gamma module (see `gamma.py` file below). Note that _scaling time to be small_ is really important to stability, example: https://github.com/CamDavidsonPilon/lifelines/pull/766/files#diff-2ba6fa67c351fd9251fde3426fbcae29R45

cristobaltapia: Wow, this is great! Thanks! I will try this tomorrow :D

cristobaltapia:
This works perfectly for me! :)

PR #797, opened 2019-08-04
fix typos

labarba:
bonus: removed a couple of "there are" "there is" syntactic expletives


labarba:
Re: https://github.com/openjournals/joss-reviews/issues/1317

CamDavidsonPilon:
🙌 thanks for the support!

PR #834, opened 2019-09-17
kaplan_meier: Change lambda -> functools.partial()

AbdealiJK:
Lambda is slightly less performant and is also not picklable.
use the functools.partial instead that is the more robust way to
manage partial functions and functions using default arguments.

### DIFF
--- a/lifelines/fitters/kaplan_meier_fitter.py
+++ b/lifelines/fitters/kaplan_meier_fitter.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import functools
 import warnings
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


AbdealiJK:
I mainly created the PR to check the tests in Travis, my local pytest seems to be failing, so I was getting confused.

@CamDavidsonPilon from our discussion in https://github.com/CamDavidsonPilon/lifelines/issues/188

The functools.partial is a better approach than lambda in cases where you are simply trying to alias a function with default arguments for ease of use / api compatibility. Lambda confuses the python interpreter as lambda are a blob (of lines) and functools.partial maintains information about which function is aliases, which arguments are default, etc.

With this fix, the KaplanMeier is picklable. I have yet to understand where the other error: `'unary_to_nary.<locals>.nary_operator.<locals>.nary_f'` is coming from. I could not find any place where `unary_to_nary` is defined in the lifelines module. Any pointers are appreciated.

CamDavidsonPilon:
❤️ that you addressed the improvement in KaplanMeierFitter, none-the-less! 

Thanks for helping out!

CamDavidsonPilon:
It looks like my recent changes have turned CI very red, please ignore those. Merging!

PR #843, opened 2019-10-04
sklearn_adapter: Better error message for invalid X

AbdealiJK:
X needs to be a dataframe currently. Print the type of X in the
exception message so it's clearer to users what is not supported.

Thought it would be a good idea, as sometimes I use this with other libraries that try to make pipelines with sklearn regressors - and Im just not sure what those other libraries are passing into lifelines (sklearn supports more inputs than lifelines - for example, sklearn supports numpy arrays)


CamDavidsonPilon:
👍 I like this. Thanks for the contribution @AbdealiJK!

PR #845, opened 2019-10-05
Fix pickle issues

AbdealiJK:
For univariate models: `_hazard` is a property in the object, but actually a function - By making it a function, it becomes a part of the class definition and hence the pickle does not consider it as a property that needs to be serialized.

From my manual testing, these fitters can now be pickled:
 - LogLogisticFitter
 - WeibullFitter
 - GeneralizedGammaFitter
Also, removed the xfail from the pickle and added joblib test cases in Univariate.

EDIT: I did some more digging and figured out the issue for the Regression models too.
For regression models: `_neg_likelihood_with_penalty_function` is a property in the object, but actually a function. By making it a functools.partial, pickle will know how to pickle it. Also, it needs the `unflatten` function which is an autograd function which cannot be pickled. So, we save the data to recreate it with autograd instead of saving the autograd function itself.

From my manual testing, these fitters can now be pickled:
 - WeibullAFTFitter
 - LogNormalAFTFitter
 - LogLogisticAFTFitter
And I also added pickle test cases for this, and removed the xfail from joblib test case for Regression models

Fixes https://github.com/CamDavidsonPilon/lifelines/issues/188

### DIFF
--- a/lifelines/fitters/__init__.py
+++ b/lifelines/fitters/__init__.py
@@ -301,9 +301,6 @@ class ParametericUnivariateFitter(UnivariateFitter):
     def __init__(self, *args, **kwargs):
         super(ParametericUnivariateFitter, self).__init__(*args, **kwargs)
         self._estimate_name = "cumulative_hazard_"
-        if not hasattr(self, "_hazard"):
-            # pylint: disable=no-value-for-parameter,unexpected-keyword-arg
-            self._hazard = egrad(self._cumulative_hazard, argnum=1)
         if not hasattr(self, "_bounds"):
             self._bounds = [(0.0, None)] * len(self._fitted_parameter_names)
         self._bounds = list(self._buffer_bounds(self._bounds))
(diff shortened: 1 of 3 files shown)


CamDavidsonPilon:
This is A+++++ 🙌 

I'm going to point this PR at my release branch and get it out soon. I'd like to do some local testing with it too.

AbdealiJK:
Sounds great, thanks!

torresxavier:
Hi @AbdealiJK !!! good post. I have not been able to save/serialize my WeibullAFTFitter. It keeps saying "" Sorry, pickling not yet supported. See pydata/patsy#26 if you want to help." I have been able to save it with a CoxPHFitter, but not with WeibullAFTFitter. And this is the one wich shows best results. 
Can you please provide the code to save/serialize the WeibullAFTFitter??
So many thanks

CamDavidsonPilon:
Commented in #1138

PR #899, opened 2019-12-11
Added import pandas to fix print_summary

millengustavo:
-  **Bug fix**
Fix NameError: name 'pd' is not defined from .print_summary() method.


CamDavidsonPilon:
Hi @millengustavo! I actually just fixed this bug this morning, and released a new version with the fix in v0.23.3.

Issue #900, opened 2019-12-15
Source installation doesn't work due to `reqs`


AbdealiJK:
When doing a source installation from pip:
`pip install --no-binary :all: lifelines`
it fails with:
```
  Downloading https://files.pythonhosted.org/packages/cd/7e/5787389141567ef38b94f1d06922d010644e32b55e471a13f80b3e1d2b2c/lifelines-0.23.3.tar.gz (321kB)
  Saved ./pypi/downloads/lifelines-0.23.3.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/g8/blwgm8mx7dng2pflkt7rwy0m0000gn/T/pip-download-kntrfzgd/lifelines/setup.py", line 8, in <module>
        with open("reqs/base-requirements.txt") as f:
    FileNotFoundError: [Errno 2] No such file or directory: 'reqs/base-requirements.txt'
```
The `reqs` folder isnt present in lifelines

CamDavidsonPilon:
oooo this might be the solution to #852!

CamDavidsonPilon:
v0.23.4 was pushed to PyPI

PR #901, opened 2019-12-15
MANIFEST: Add "reqs"

AbdealiJK:
If not added, source installations created with `sdist`
don't work.

Fixes https://github.com/CamDavidsonPilon/lifelines/issues/900


AbdealiJK:
@CamDavidsonPilon would it be possible to push a bugfix version into pypi for this issue ?

CamDavidsonPilon:
@AbdealiJK thanks for this PR - and yes, I think it's important enough to hotfix on PyPI.

PR #943, opened 2020-02-04
Add file extension to broken link.

matthewbegun:
There's also a `py` file there but I assume this should point at the notebook. There is no directory there (404).


CamDavidsonPilon:
LGTM! Thanks!

PR #945, opened 2020-02-05
Remove nested rst formatting from contributing

bluemoo:
The formatting in the bottom sections of the ['Contributing' page on readthedocs](https://lifelines.readthedocs.io/en/latest/Contributing.html#setting-up-a-lifelines-development-environment) is broken, resulting in the display of formatting characters and full urls. The contents should instead mirror the [github contributing page](https://github.com/CamDavidsonPilon/lifelines/blob/master/.github/CONTRIBUTING.md). The problems mostly stem from the fact that (unlike markdown) rst formatting cannot be nested, so you can't have something that's both italic and a link.

This PR is an example in one possible way to restructure things such that the contents appear more in line with the github based 'contributing' page, plus a couple of minor typo fixes.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -34,8 +34,8 @@ There are 3 different ways to format your code.
    * `make lint`
 2. Call `black` directly and pass the correct line length.
    * `black . -l 120`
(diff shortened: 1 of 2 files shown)


CamDavidsonPilon:
This is great! Thanks, @bluemoo

PR #947, opened 2020-02-07
Add a generic mixture cure fitter

bluemoo:
This is a prototype of what a generic mixture cure fitter could look like. At the time of writing, it's probably too early to actually be merged.

### DIFF
--- a/lifelines/fitters/mixture_cure_fitter.py
+++ b/lifelines/fitters/mixture_cure_fitter.py
@@ -11,8 +11,8 @@ class MixtureCureFitter(ParametricUnivariateFitter):
 
         if self.CURED_FRACTION_PARAMETER_NAME in base_fitter._fitted_parameter_names:
             raise NameError(
-                f"'{self.CURED_FRACTION_PARAMETER_NAME}' in _fitted_parameter_names is a lifelines reserved word. Try something "
-                "else instead."
+                f"'{self.CURED_FRACTION_PARAMETER_NAME}' in _fitted_parameter_names is a lifelines reserved word."
+                f" Try something else instead."
             )
 
         self._fitted_parameter_names = [self.CURED_FRACTION_PARAMETER_NAME] + base_fitter._fitted_parameter_names
--- a/lifelines/fitters/mixture_cure_fitter.py
+++ b/lifelines/fitters/mixture_cure_fitter.py
@@ -28,8 +28,13 @@ class MixtureCureFitter(ParametricUnivariateFitter):
 
     def _cumulative_hazard(self, params, times):
         c = params[0]
-        sf = self._base_fitter._survival_function(params[1:], times)
-        return -anp.log(c + (1 - c) * sf)
+        base_survival_function = self._base_fitter._survival_function(params[1:], times)
+        return -anp.log(c + (1 - c) * base_survival_function)
+
+    def _survival_function(self, params, times):
+        c = params[0]
+        base_cumulative_hazard = self._base_fitter._cumulative_hazard(params[1:], times)
+        return c + (1 - c) * (anp.exp(-base_cumulative_hazard))
 
     def _create_initial_point(self, Ts, E, *args):
         base_point = self._base_fitter._create_initial_point(Ts, E, *args)
--- a/lifelines/fitters/mixture_cure_fitter.py
+++ b/lifelines/fitters/mixture_cure_fitter.py
@@ -1,9 +1,8 @@
 # -*- coding: utf-8 -*-
 import autograd.numpy as anp
-from autograd.scipy.special import expit
 from lifelines.fitters import ParametricUnivariateFitter
 
-# What should the name of the fitter actually be?
+
 class MixtureCureFitter(ParametricUnivariateFitter):
     CURED_FRACTION_PARAMETER_NAME = "cured_fraction_"
 
--- a/lifelines/fitters/mixture_cure_fitter.py
+++ b/lifelines/fitters/mixture_cure_fitter.py
@@ -36,6 +36,13 @@ class MixtureCureFitter(ParametricUnivariateFitter):
         base_cumulative_hazard = self._base_fitter._cumulative_hazard(params[1:], times)
         return c + (1 - c) * (anp.exp(-base_cumulative_hazard))
 
+    def _log_hazard(self, params, times):
+        c = params[0]
+        base_survival_function = self._base_fitter._survival_function(params[1:], times)
+        base_log_hazard = self._base_fitter._log_hazard(params[1:], times)
+        cumulative_hazard = -anp.log(c + (1 - c) * base_survival_function)
+        return anp.log(1 - c) + anp.log(base_survival_function) + base_log_hazard + cumulative_hazard
+
     def _create_initial_point(self, Ts, E, *args):
         base_point = self._base_fitter._create_initial_point(Ts, E, *args)
         return anp.array([0] + list(base_point))
--- a/lifelines/fitters/mixture_cure_fitter.py
+++ b/lifelines/fitters/mixture_cure_fitter.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+import autograd.numpy as anp
+from autograd.scipy.special import expit
+from lifelines.fitters import ParametricUnivariateFitter
+
+# What should the name of the fitter actually be?
+class MixtureCureFitter(ParametricUnivariateFitter):
+    def __init__(self, base_fitter, *args, **kwargs):
+        self._base_fitter = base_fitter
+
+        # What should the variable that we use be called?
+        # Should raise an exception if that variable is in the base fitted parameter names
+
+        self._fitted_parameter_names = ["cured_"] + base_fitter._fitted_parameter_names
+        self._bounds = [(None, None)] + base_fitter._bounds
+        super().__init__(*args, **kwargs)
+
+    def _fit(self, *args, **kwargs):
+        self._base_fitter._censoring_type = self._censoring_type
+        result = super()._fit(*args, **kwargs)
+        for param_name, fitted_value in zip(self._fitted_parameter_names[1:], self._fitted_parameters_[1:]):
+            setattr(self._base_fitter, param_name, fitted_value)
+        return result
+
+    def _cumulative_hazard(self, params, times):
+        c = expit(params[0])
+        sf = self._base_fitter._survival_function(params[1:], times)
+        return -anp.log(c + (1 - c) * sf)
+
+    def _create_initial_point(self, Ts, E, *args):
+        base_point = self._base_fitter._create_initial_point(Ts, E, *args)
+        return anp.array([0] + list(base_point))
+
+    def percentile(self, p):
+        c = expit(self.cured_)
+
+        if p <= c:
+            raise ValueError("Percentile must be larger than the cure fraction")
+
+        non_cure_p = (p - c) / (1 - c)
+        return self._base_fitter.percentile(non_cure_p)
--- a/lifelines/fitters/mixture_cure_fitter.py
+++ b/lifelines/fitters/mixture_cure_fitter.py
@@ -35,7 +35,7 @@ class MixtureCureFitter(ParametricUnivariateFitter):
         c = self.cured_fraction_
 
         if p <= c:
-            raise ValueError("Percentile must be larger than the cure fraction")
+            return anp.inf
 
         non_cure_p = (p - c) / (1 - c)
         return self._base_fitter.percentile(non_cure_p)
--- a/lifelines/fitters/mixture_cure_fitter.py
+++ b/lifelines/fitters/mixture_cure_fitter.py
@@ -5,11 +5,16 @@ from lifelines.fitters import ParametricUnivariateFitter
 
 # What should the name of the fitter actually be?
 class MixtureCureFitter(ParametricUnivariateFitter):
+    CURED_FRACTION_PARAMETER_NAME = "cured_fraction_"
+
     def __init__(self, base_fitter, *args, **kwargs):
         self._base_fitter = base_fitter
 
-        # What should the variable that we use be called?
-        # Should raise an exception if that variable is in the base fitted parameter names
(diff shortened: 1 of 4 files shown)


lifelines/tests/test_estimation.py
```diff
@@ -946,7 +947,7 @@ def test_weibull_fit_returns_float_timelines(self):
         npt.assert_array_equal(wf.timeline, T)
         npt.assert_array_equal(wf.survival_function_.index.values, T)
 
-    def test_weibull_model_does_not_except_negative_or_zero_values(self):
+    def test_weibull_model_does_not_accept_negative_or_zero_values(self):
```
CamDavidsonPilon: lol thanks

lifelines/fitters/mixture_cure_fitter.py
```diff
@@ -0,0 +1,41 @@
...
+        self._base_fitter._censoring_type = self._censoring_type
+        result = super()._fit(*args, **kwargs)
+        for param_name, fitted_value in zip(self._fitted_parameter_names[1:], self._fitted_parameters_[1:]):
+            setattr(self._base_fitter, param_name, fitted_value)
+        return result
+
+    def _cumulative_hazard(self, params, times):
+        c = expit(params[0])
+        sf = self._base_fitter._survival_function(params[1:], times)
+        return -anp.log(c + (1 - c) * sf)
```
CamDavidsonPilon: 👍

lifelines/fitters/mixture_cure_fitter.py
```diff
@@ -0,0 +1,41 @@
...
+
+# What should the name of the fitter actually be?
+class MixtureCureFitter(ParametricUnivariateFitter):
+    def __init__(self, base_fitter, *args, **kwargs):
+        self._base_fitter = base_fitter
+
+        # What should the variable that we use be called?
+        # Should raise an exception if that variable is in the base fitted parameter names
+
+        self._fitted_parameter_names = ["cured_"] + base_fitter._fitted_parameter_names
```
CamDavidsonPilon: Some suggestions:
`"logit_cured"`, `"logit_cured_prob"`.


Any reason to use the model the logit? The alternative is to model the probability and use bounds `(0, 1)` - this gives a much easier interpretation for end users.

lifelines/fitters/mixture_cure_fitter.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def _create_initial_point(self, Ts, E, *args):
+        base_point = self._base_fitter._create_initial_point(Ts, E, *args)
+        return anp.array([0] + list(base_point))
+
+    def percentile(self, p):
+        c = expit(self.cured_)
+
+        if p <= c:
+            raise ValueError("Percentile must be larger than the cure fraction")
```
CamDavidsonPilon: is it better to return `inf`?

lifelines/fitters/mixture_cure_fitter.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+import autograd.numpy as anp
+from autograd.scipy.special import expit
+from lifelines.fitters import ParametricUnivariateFitter
+
+# What should the name of the fitter actually be?
+class MixtureCureFitter(ParametricUnivariateFitter):
```
CamDavidsonPilon: Part of me wants to call this `"Wrapper"`, to distinguish its unique API from traditional `Fitters` - but it is still a `Fitter`. Not sure..

lifelines/tests/test_estimation.py
```diff
@@ -4841,3 +4842,43 @@ def test_ci_calculation_against_sas(self, fitter, duration, event_observed):
...
+
+
+class TestMixtureCureFitter:
+    def test_exponential_data_produces_correct_inference_for_both_cure_and_non_cure_fractions(self):
+        N = 1000000
+        scale = 5
+        T = np.random.exponential(scale, size=N)
+        observed = np.ones(N, dtype=bool)
+
+        # Censor the data at time = 6
```
CamDavidsonPilon: 8?

lifelines/tests/test_estimation.py
```diff
@@ -4841,3 +4842,43 @@ def test_ci_calculation_against_sas(self, fitter, duration, event_observed):
...
+
+        # Add in some 'cured' samples, to make it 20% cured
+        C = int(N / 4)
+        T = np.concatenate([T, last_observation_time * np.ones(C)])
+        observed = np.concatenate([observed, np.zeros(C, dtype=bool)])
+
+        fitter = MixtureCureFitter(base_fitter=ExponentialFitter())
+        fitter.fit(T, event_observed=observed)
+
+        assert abs(expit(fitter.cured_) - 0.2) < 0.01
```
CamDavidsonPilon: 👍

CamDavidsonPilon:
I'm digging this. I left some comments. Naming is still up in the air. Another testsuite to add this to is the list in `TestUnivariateFitters`

lifelines/fitters/mixture_cure_fitter.py
bluemoo: I have a data set for which I have built a mixture cure model as a ParametricRegressionFitter with several covariates. If I just use the identify link, then the fit does not converge. Instead I use logit and it works well. I assumed it would be the same here, but I'm actually unable to replicate using this univariate fitter. I've pushed a new commit for you to look at which models things as you've described.

lifelines/fitters/mixture_cure_fitter.py
bluemoo: Sounds better to me - done.

lifelines/fitters/mixture_cure_fitter.py
bluemoo: I was thinking of the way sklearn ensembles work when building this. For instance, the [AdaBoostClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.AdaBoostClassifier.html#sklearn.ensemble.AdaBoostClassifier) takes a base_estimator and is still called a Classifier. Thoughts?

lifelines/tests/test_estimation.py
bluemoo: This is a great example of why tests are better documentation than comments. Harder to get out of date. In this case... instantly. Anyway, fix pushed.

bluemoo:
Should the _KNOWN_MODEL flag be derived from the base fitter? For instance, if I use the built in ExponentialFitter, then _KNOWM_MODEL = True. If I supply my own, then it's False?

bluemoo:
I've added the model to the TestUnivariateFitters list, but had to work around that fact that currently base_fitter is a required field. In the commit I defined a subclass for testing that forced the base_fitter to be an ExponentialFitter. Instead we could just allow base_fitter to be optional, and use a default fitter if none is specified. Is one approach better than the other?

lifelines/fitters/mixture_cure_fitter.py
CamDavidsonPilon: I've seen this too - for regression data, using unbounded parameterizations works much better. But for univariate fitters, it's less important. I think the code without link is better for this :+1:

lifelines/fitters/mixture_cure_fitter.py
CamDavidsonPilon: good with me then!

lifelines/tests/test_estimation.py
```diff
@@ -330,10 +331,16 @@ def test_print_summary(self, sample_lifetimes, known_parametric_univariate_fitte
...
+class ExponentialBaseMixtureCureFitter(MixtureCureFitter):
+    def __init__(self):
+        super().__init__(base_fitter=ExponentialFitter())
+
+
 class TestUnivariateFitters:
     @pytest.fixture
     def univariate_fitters(self):
         return [
+            ExponentialBaseMixtureCureFitter,
```
CamDavidsonPilon: this is fine, but can't you also do:

```
return 
  [ 
    MixtureCureFitter(ExponentialFitter()),
   ...
```

lifelines/tests/test_estimation.py
```diff
@@ -4841,3 +4848,42 @@ def test_ci_calculation_against_sas(self, fitter, duration, event_observed):
 
         fitter.fit(duration, event_observed, event_of_interest=2)
         npt.assert_allclose(ci_from_sas, np.array(fitter.confidence_interval_))
+
+
+class CureModelExponential(ParametricUnivariateFitter):
```
CamDavidsonPilon: not used?

lifelines/tests/test_estimation.py
bluemoo: The univariate_fitters fixture returns a list of callables, rather than class instances, so that wouldn't work directly. You see this in the later tests which need more than once instance of the class, and do things like:
```
        for fitter in univariate_fitters:
            f1 = fitter()
            f2 = fitter()
```
However, your suggestion made me realize that a simple lambda might actually be less weird than that extra class I had. So I've committed an example of what that might look like for you. It also has a side effect of fixing two tests that I broke :)

lifelines/tests/test_estimation.py
bluemoo: Oops, nice catch.

lifelines/tests/test_estimation.py
```diff
@@ -331,16 +331,11 @@ def test_print_summary(self, sample_lifetimes, known_parametric_univariate_fitte
...
-    def __init__(self):
-        super().__init__(base_fitter=ExponentialFitter())
-
-
 class TestUnivariateFitters:
     @pytest.fixture
     def univariate_fitters(self):
         return [
-            ExponentialBaseMixtureCureFitter,
+            lambda *args, **kwargs: MixtureCureFitter(base_fitter=ExponentialFitter(), *args, **kwargs),
```
CamDavidsonPilon: oooooo I see - yea this is awkward. I'll fix this better in the future.

bluemoo:
@CamDavidsonPilon Is there anything else you'd like to have added / removed / changed? I think all of the important functions are defined explicitly in terms of the base_fitter, but it's certainly possible I missed something. Should we do anything special with the KNOWN_MODEL flag? If not, then I think the next step is documentation.

lifelines/fitters/mixture_cure_fitter.py
```diff
@@ -0,0 +1,57 @@
...
+    def _log_hazard(self, params, times):
+        c = params[0]
+        base_survival_function = self._base_fitter._survival_function(params[1:], times)
+        base_log_hazard = self._base_fitter._log_hazard(params[1:], times)
+        cumulative_hazard = -anp.log(c + (1 - c) * base_survival_function)
+        return anp.log(1 - c) + anp.log(base_survival_function) + base_log_hazard + cumulative_hazard
+
+    def _create_initial_point(self, Ts, E, *args):
+        base_point = self._base_fitter._create_initial_point(Ts, E, *args)
+        return anp.array([0] + list(base_point))
```
CamDavidsonPilon: 0.5?

lifelines/fitters/mixture_cure_fitter.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def percentile(self, p):
+        c = self.cured_fraction_
+
+        if p <= c:
+            return anp.inf
+
+        non_cure_p = (p - c) / (1 - c)
+        return self._base_fitter.percentile(non_cure_p)
```
CamDavidsonPilon: :smart:

CamDavidsonPilon:
I think `KNOWN_MODEL=True` is correct for this. 

Overall, code looks good to me! Thanks for thinking through the details and dissecting my internal lifelines code. 

W.r.t documentation, all the current cure documentation is tied to Custom models under "Time-lagged conversion rates and cure models". I would be happy to see changes here introduce your new feature. Are you interested in doing that in this PR?

bluemoo:
I've set the known model flag and added it to the known model parametric univariate fitters list, so that the plotting tests run it. With that, I think at least a first version of this fitter itself is done. I'd be happy to either write up some documentation in this PR or a separate one, as you think is appropriate. Obviously, there's not much utility in having the fitter in the codebase without documentation to point people at it, but it might make things simpler to have a new documentation-only PR. Also, I noticed that CI tests are failing. Is there something I'm missing that will get them to pass?

bluemoo:
I think that there are three changes to be made for documentation:

- Add it to the "Time-lagged conversion rates and cure models" section.
- Add class docstring
- Add to the list in docs/lifelines.fitters.rst

Does that sound right?

CamDavidsonPilon:
> Also, I noticed that CI tests are failing. Is there something I'm missing that will get them to pass?

ignore it! Something I'm working on in #927. 

Your list sounds about right. Would you prefer to start a new PR for the documentation? I'm fine with that.

bluemoo:
It felt wrong for there to be no documentation at all, so I added a docstring to the class and put the MixtureCureFitter into the API Documentation. If you're happy with it, I think this PR is minimally complete now and could be merged! Either way, I'd love feedback regarding any improvements you think could be made to the docstring, and/or ideas for specifically what to discuss in the cure models section elsewhere.

lifelines/fitters/mixture_cure_fitter.py
```diff
@@ -0,0 +1,122 @@
...
+    alpha: float, optional (default=0.05)
+        the level in the confidence intervals.
+
+    Important
+    ----------
+    The **base_fitter** instance is used to describe the non-cure portion of the population, but is not actually
+    fit to the data. Some internal properties are modified, and it should not be used for any other purpose after
+    passing it to the constructor of this class.
+
+    Examples
```
CamDavidsonPilon: 👍

CamDavidsonPilon:
Great job on this, @bluemoo! 

I'll merge this now. 

For more documentation, I think adding a section in "Time-lagged conversion rates and cure models" makes sense , something like (paraphrasing):

 - "Here's how to create a cure model using lifelines custom semantics: [code]"
 - "This operation is so common, we've introduced a wrapper to make it easier. The class `MixtureCureFitter` is used like this: [code]"

We can also add it to the docs / images in https://lifelines.readthedocs.io/en/latest/Survival%20analysis%20with%20lifelines.html#other-parametric-models-exponential-log-logistic-log-normal-and-splines.

PR #950, opened 2020-02-10
Add Python 3.9 compatibility

tirkarthi:
* Import collections from ABC instead of collections since it was removed in Python 3.9.
* Use raw string to fix a deprecation warning regrading invalid escape sequences.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ python:
    - "3.5"
    - "3.6"
    - "3.7"
+   - "3.8"
+   - "3.9-dev"
 env:
   matrix:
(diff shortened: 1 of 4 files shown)


CamDavidsonPilon:
Excellent, thanks! Can you also update the setup.py with the new CLASSIFIERS? I'd do it myself, but I worry about forgetting.

tirkarthi:
Added 3.8 and 3.9 to travis and setup.py.

CamDavidsonPilon:
lgtm, thanks!

tirkarthi:
Thanks :)

PR #951, opened 2020-02-11
Updated printer.py (LaTeX table with proper decimals)

arturomoncadatorres:
Printing a CPH coefficient table is done using [`.to_latex()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_latex.html) (with no parameters). However, this means that the printed number have all decimals. With this fix, the number of printed decimals corresponds to the defined parameter `decimals`.

### DIFF
--- a/lifelines/utils/printer.py
+++ b/lifelines/utils/printer.py
@@ -47,7 +47,8 @@ class Printer:
         print(self.to_latex())
 
     def to_latex(self):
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
Yes! Nice catch, thanks @arturomoncadatorres!

PR #977, opened 2020-03-15
DOC: little copyedit to survival regression docs

aflaxman:
Thank you for this clear and usable package and documentation!


CamDavidsonPilon:
oh lovely change! Thanks @aflaxman 

PS Your blog was an early influence on my statistical thinking :)

aflaxman:
Last fall I got a bunch of my coworkers to work through your PP4Hackers book, so what goes around comes around. Thanks again!

Issue #995, opened 2020-04-20
plot_lifetimes() / Multicenter Aids Cohort Study Example


jan-marchant:
Hi there, thanks for this great resource. I think there's a small issue with the plot_lifetimes() example using Multicenter Aids Cohort Study data given here:

https://lifelines.readthedocs.io/en/latest/Survival%20analysis%20with%20lifelines.html

With recent Pandas the example fails with a KeyError. This is because plot_lifetimes() can only properly handle dataframes with a sequential index starting at 0, and the example data starts at index 1. This doesn't error with Pandas < 1.0 which instead silently drops rows with mismatched indices (as described here: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#deprecate-loc-reindex-listlike')

The solution that works for me is to load the example data with:

df = load_multicenter_aids_cohort_study().reset_index(drop=True)

but I expect a better solution is to reset the index inside plot_lifetimes() so that whatever data is passed is handled properly.

Cheers,
Jan

sean-reed:
Additionally, it doesn't sort by duration correctly when durations argument is a Panda DataSeries. Example:
```
from lifelines.plotting import plot_lifetimes
import pandas as pd

durations = [10, 15, 5, 25, 20]
event_observed = [1, 0, 0, 1, 1] 

df = pd.DataFrame({'durations': durations, 'event_observed': event_observed})

plot_lifetimes(df['durations'], df['event_observed'])
```

sean-reed:
Related to this, would it be better to have the plot bars be labelled by the corresponding indices from the Pandas series instead of 0 to 1? It does this already with the default index but if the index is something like ["Item 1", "Item 2", "Item 3"] instead, then it seems better to have this displayed next to the bars on the y-axis instead of 0, 1, 2. Now would be the ideal time to implement this, since any code using plot_lifetimes with a non-default index currently fails due to the issue raised above and for anything else the output would remain the same. Could also add as an optional parameter like label_by_index=True. Thoughts?

CamDavidsonPilon:
Whoops, sorry @jan-marchant, seems I missed this issue. But luckily @sean-reed is here to help in https://github.com/CamDavidsonPilon/lifelines/pull/1003

@sean-reed your suggestion of using index values is great. Want to implement that in your PR as well?

sean-reed:
Great, I've pushed the update.

PR #1003, opened 2020-04-30
Fixes issue #995.

sean-reed:
plot_lifetimes now handles Pandas Series correctly.

### DIFF
--- a/lifelines/plotting.py
+++ b/lifelines/plotting.py
@@ -478,6 +478,17 @@ def plot_lifetimes(
         ax = plot_lifetimes(T.loc[:50], event_observed=E.loc[:50])
 
     """
+
+    def iloc(x, i):
+        """
+        Returns the item at index i or items at indices i from x,
+        where x is a numpy array or pd.Series.
+        """
+        try:
+            return x.iloc[i]
+        except AttributeError:
+            return x[i]
+
     if ax is None:
         ax = plt.gca()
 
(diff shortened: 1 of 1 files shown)


lifelines/plotting.py
```diff
@@ -441,6 +441,7 @@ def plot_lifetimes(
     sort_by_duration=True,
     event_observed_color="#A60628",
     event_censored_color="#348ABD",
+    label_durations=False,
```
CamDavidsonPilon, line 444: Thoughts on always using the index values if provided a pd.Series? By default, the index is 0, 1, 2, ... which is identical to current behaviour. I feel: if a user provided a pd.Series with non-trivial index values, they expect to see them.

sean-reed: The only issue I see with always labelling with values from the index is that the labelling becomes cluttered if there are lot of durations to plot, e.g. with the waltons example in the doc string.

CamDavidsonPilon: Hm, that might be exception. Do you have a quick screenshot of how cluttered it could get? 

(Also, walton's data doesn't have an non-trivial index, right?)

sean-reed: I see what you are saying, I've updated the function to only add the labels if durations is a pd.Series with non-default index (had to add the check before the durations sort since that rewrites the index leaving no way to tell if it was non-default).

CamDavidsonPilon:
lgtm! Thanks for the contribution @sean-reed!

PR #1006, opened 2020-05-01
Added plotting of lifetimes from interval censored data

sean-reed:
Added plot_interval_censored_lifetimes function - similar to plot_lifetimes but for interval censored data. Example from doc string:
lifetimes_plot

### DIFF
--- a/lifelines/plotting.py
+++ b/lifelines/plotting.py
@@ -12,7 +12,15 @@ import pandas as pd
 from lifelines.utils import coalesce, CensoringType
 
 
-__all__ = ["add_at_risk_counts", "plot_lifetimes", "qq_plot", "cdf_plot", "rmst_plot", "loglogs_plot"]
+__all__ = [
+    "add_at_risk_counts",
+    "plot_lifetimes",
+    "plot_interval_censored_lifetimes",
+    "qq_plot",
+    "cdf_plot",
+    "rmst_plot",
+    "loglogs_plot",
+]
 
 
 def _iloc(x, i):
@@ -190,9 +198,7 @@ def rmst_plot(model, model2=None, t=np.inf, ax=None, text_position=None, **plot_
         )
 
         ax.text(
-            text_position[0],
-            text_position[1],
-            "RMST(%s) -\n   RMST(%s)=%.3f" % (model._label, model2._label, rmst - rmst2),
+            text_position[0], text_position[1], "RMST(%s) -\n   RMST(%s)=%.3f" % (model._label, model2._label, rmst - rmst2),
         )  # dynamically pick this.
     else:
         rmst = restricted_mean_survival_time(model, t=t)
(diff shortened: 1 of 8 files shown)


lifelines/plotting.py
```diff
@@ -433,6 +434,113 @@ def add_at_risk_counts(*fitters, **kwargs):
...
+    .. code:: python
+
+        import pandas as pd
+        import numpy as np
+        from lifelines.plotting import plot_interval_censored_lifetimes
+        df = pd.DataFrame({'lb':[20,15,30, 10, 20, 30], 'ub':[25, 15, np.infty, 20, 20, np.infty]})
+        ax = plot_interval_censored_lifetimes(lower_bound=df['lb'], upper_bound=df['ub'])
+    """
+
+    def iloc(x, i):
```
CamDavidsonPilon, line 485: I think we should pull this out into a top level function so both `plot_..._lifetimes` function can use it.

lifelines/plotting.py
```diff
@@ -527,7 +635,7 @@ def iloc(x, i):
         ax.set_yticks(range(0, N))
         ax.set_yticklabels(durations.index)
     else:
-        ax.set_yticks([])
+        ax.yaxis.set_major_locator(MaxNLocator(integer=True))
```
CamDavidsonPilon, line 638: Can you explain what `MaxNLocator` does?

lifelines/plotting.py
```diff
@@ -433,6 +434,113 @@ def add_at_risk_counts(*fitters, **kwargs):
...
+        entry = np.zeros(N)
+
+    for i in range(N):
+        if np.isposinf(iloc(upper_bound, i)):
+            c = event_right_censored_color
+            ax.hlines(i, iloc(entry, i), iloc(lower_bound, i), color=c, lw=1.5)
+        else:
+            c = event_observed_color
+            ax.hlines(i, iloc(entry, i), iloc(upper_bound, i), color=c, lw=1.5)
+            if iloc(lower_bound, i) == iloc(upper_bound, i):
```
CamDavidsonPilon, line 524: 👍

CamDavidsonPilon:
This is a really slick solution and viz! Left a few single comments.

lifelines/plotting.py
sean-reed: It just ensures that any y-axis labels are at integer values, otherwise if durations has just a few values (seems to happen with about 20 or less):
```
from lifelines.plotting import plot_lifetimes
import numpy as np
a = np.asarray([20.0, 53.0, 50.0])
plot_lifetimes(a)
```
it results in something like this:
plot1
instead of:
plot2

lifelines/plotting.py
sean-reed: Yes, good point, I've moved it.

CamDavidsonPilon:
Looks great, @sean-reed! Thanks again!

sean-reed:
Thanks!

Issue #452, opened 2018-05-05
Log-rank test and Breslow (Generalized Wilcoxon)


yong27:
Thanks for great project!

I'm fan of Lifelines. I'm usually doing log-rank test for two groups. 

In SPSS, there are 3 kinds of test (Log Rank, Breslow (Generalized Wilcoxon), Tarone-Ware) and each description is as follows.

* Log Rank: A test for comparing the equality of survival distributions. All time points are weighted equally in this test.
* Breslow: A test for comparing the equality of survival distributions. Time points are weighted by the number of cases at risk at each time points.
* Tarone-Ware: A test for comparing the equality of survival distributions. Time points are weighted by the square root of the number of cases at risk at each time point.

It seems to easy to implement other tests because of good code quality of lifelines. But I failed to get the same p-value with SPSS.

How can I implement these tests?

Thank you.

CamDavidsonPilon:
Hey @yong27, thanks for the issue. 

Are you saying that the `utils. logrank_test` in lifelines produced different results from SPSS? As for the other two tests, no, those don't exist yet. Those seem like good additions though.

sean-reed:
I had a go at adding the Wilcoxon and Tarone-Ware test stats: https://github.com/sean-reed/lifelines/blob/bb93312c4462c5a808118a6f1b79ea76aab2a9f7/lifelines/statistics.py#L649

Not much to the implementation of the tests themselves since just a case of applying weightings in the numerator and covariance matrix computations within the existing log-rank statistic implementation - mostly a refactoring to avoid duplicated code between the test statsistics. I still need to add some unit tests but were correct on on a few test examples I tried...

sean-reed:
There are a few other popular variations on the log-rank test that could also be added such as Peto and Flemington-Harrington.

CamDavidsonPilon:
:+1: that's a nice refactor, @sean-reed - happy to merge changes like those in

sean-reed:
Great, I will add the other log-rank test variants and the unit tests etc. soon.

PR #1030, opened 2020-05-24
Added Wilcoxon, Tarone-Ware and Peto log-rank alternative tests

sean-reed:
See issue #452

### DIFF
--- a/docs/Examples.rst
+++ b/docs/Examples.rst
@@ -93,6 +93,56 @@ hypothesis that all the populations have the same "death" generation process).
     ---
     """
 
+The logrank test statistic is calculated from the differences between the observed deaths for a group and expected
+deaths, under the null hypothesis that all groups share the same survival curve, summed across all ordered death times.
+It therefore weights differences between the survival curves equally at each death time, resulting in maximum power
+when the assumption of proportional hazards is true. To test for early or late differences in survival between
+groups, a weighted logrank test that are more sensitive to non-proportional hazards might be a better choice.
+
+Four types of weighted logrank test are currently available in lifelines through the ``weightings`` argument:
+the Wilcoxon (``weightings='wilcoxon'``), Tarone-Ware (``weightings='tarone-ware'``), Peto (``weightings='peto'``)
+and Fleming-Harrington (``weightings='fleming-harrington'``) tests.
+The following weightings are applied at the ith ordered failure time, :math:`t_{i}`:
+
+    .. math:: \text{Wilcoxon:}\quad n_i
+    .. math:: \text{Tarone-Ware:}\quad \sqrt{n_i}
+    .. math:: \text{Peto:}\quad \bar{S}(t_i)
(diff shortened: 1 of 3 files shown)


lifelines/tests/test_statistics.py
```diff
@@ -103,6 +103,48 @@ def test_rank_test_output_against_R_with_censorship():
     assert abs(result.test_statistic - r_stat) < 10e-6
 
 
+def test_wilcoxon_weighted_logrank_on_leukemia_dataset():
```
CamDavidsonPilon, line 106: 👍 nice

lifelines/statistics.py
```diff
@@ -646,6 +675,18 @@ def multivariate_logrank_test(
     t_0: float, optional (default=-1)
         the period under observation, -1 for all time.
 
+    weightings: str, optional
+        apply a weighted logrank test: options are "w" for Wilcoxon (also known as Breslow), "tw"
```
CamDavidsonPilon, line 679: I feel these should be more explicit. Ex: `weightings="wilcoxon"` vs `weights="w"`. The latter feels like an R thing, and the former a Python thing (at least, how I want Python to be). Thoughts?

CamDavidsonPilon:
Love it, great addition. Can you comment about _why_ one would want to use different weightings?

lifelines/statistics.py
sean-reed: Yes, that seems better.

sean-reed:
The weighted tests put emphasis on the beginning of the survival curve (e.g. Wilcoxon where weight is based on remaining number at risk), increasing the power of the test when it's expected the difference in survival is early on,  whereas the standard log-rank test has maximum power for proportional hazards. For example, certain drugs reduce death rates early on (e.g. first 10 years over 20 year follow up) but have less effect on the death rate for those that survive beyond a certain point (e.g. between years 10 and 20). Another example might be an online store testing whether including a discount voucher with a customer's order influences the time to them placing their next order. You might expect it to influence the rate of re-ordering for the first few weeks after receiving the voucher but have a diminishing influence later on, so a weighted test may be more appropriate. Should add that the choice of whether to weight the test should be made in advance to avoid introducing bias. I can add something to the docs on this at some point.

CamDavidsonPilon:
>  I can add something to the docs on this at some point.

What you wrote is great 👍

CamDavidsonPilon:
PR is great - did you want to add documentation to `/docs` or do you want me to add later?

sean-reed:
Thanks, leave it to me I'll add it.

sean-reed:
Added some docs and also the Fleming-Harrington test (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6047314/#bib4) that allows testing for late differences in survival.

docs/Examples.rst
```diff
@@ -93,6 +93,56 @@ hypothesis that all the populations have the same "death" generation process).
     ---
     """
 
+The logrank test statistic is calculated from the differences between the observed deaths for a group and expected
```
CamDavidsonPilon, line 96: Awesome!

PR #1133, opened 2020-08-31
Update Survival Regression.rst

ckarjadi:
extra 'model' in the sentence "The idea behind Cox's proportional hazard **model model** is that the...over time."


CamDavidsonPilon:
thanks @ckarjadi!

PR #1161, opened 2020-11-06
fix k_fold_cross_validation documentation

m-colombo:
It looked outdated, tried to update that accordingly to other documenation


CamDavidsonPilon:
ah nice, ty @m-colombo!

PR #1165, opened 2020-11-17
Silence deprecation warning

jseabold:
Adopt new pandas behavior and silence deprecation warning. This can be quite noisy in fit if there's a datetime in your DataFrame for pandas >= 1.1.0.

Cf. https://github.com/pandas-dev/pandas/pull/34798

### DIFF
--- a/lifelines/fitters/__init__.py
+++ b/lifelines/fitters/__init__.py
@@ -1250,7 +1250,7 @@ class RegressionFitter(BaseFitter):
             return v
 
         else:
-            described = df.describe(include="all")
+            described = df.describe(include="all", datetime_is_numeric=True)
             if "top" in described.index and "50%" not in described.index:
                 central_stats = described.loc["top"].copy()
             elif "50%" in described.index and "top" not in described.index:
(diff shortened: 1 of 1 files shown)


lifelines/fitters/__init__.py
```diff
@@ -1250,7 +1250,13 @@ def _compute_central_values_of_raw_training_data(self, df, strata=None, name="ba
             return v
 
         else:
-            described = df.describe(include="all")
+            from distutils.version import LooseVersion
+            if LooseVersion(pd.__version__) >= '1.1.0':
```
CamDavidsonPilon, line 1254: TIL `LooseVersion`

CamDavidsonPilon:
Nice, thanks for the PR @jseabold! Failing tests are unrelated (and the status quo 🙃)

lifelines/fitters/__init__.py
jseabold: One thing is that distutils will be deprecated in Python 3.9/3.10. Might make setuptools a runtime dependency at some point. Haven't been keeping up with this.

PR #1206, opened 2021-01-22
V0.25.8

CamDavidsonPilon:
#### 0.25.8 - 2021-01-22

Important: we dropped Patsy as our formula framework, and adopted Formulaic. Will the latter is less mature than Patsy, we feel the core capabilities are satisfactory and it provides new opportunities.

##### New features
 - Parametric models with formulas are able to be serialized now.
 - a `_scipy_callback` function is available to use in fitting algorithms.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 
 ### Questions about survival analysis?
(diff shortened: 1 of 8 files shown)


CamDavidsonPilon:
ran tests locally, all good.

PR #1238, opened 2021-03-04
if the first entry was after 0, this min would produce a nan

CamDavidsonPilon:


lifelines/plotting.py
```diff
@@ -504,7 +505,7 @@ def add_at_risk_counts(
                             lbl += ("\n" if i > 0 else "") + r"%s" % labels[int(i / n_rows)] + "\n"
 
                     l = rows_to_show[i % n_rows]
-                    l = {"At risk": "  At risk", "Censored": "Censored  ", "Events": " Events  "}.get(l)
+                    l = {"At risk": " At risk  ", "Censored": "Censored  ", "Events": " Events  "}.get(l)
```
CamDavidsonPilon, line 508: also fixes an align issue

CamDavidsonPilon:
why didn't I merge this?

Issue #1251, opened 2021-04-10
Incorrect elastic net penality in Cox Models


bnaras:
The elastic net penalty is incorrect for Cox models (lifelines version 0.25.11 and all others). The correct penalty term is
$$
\lambda\left(\alpha \sum_{j=1}^{p}\left|\beta_{j}\right|+\frac{1-\alpha}{2} \sum_{j=1}^{p} \beta_{j}^{2}\right)
$$
Luckily, this seems to affect only `cox_time_varying_fitter.py` and `coxph_fitter.py` and everywhere else it is correct. I will make a PR referring to this issue.

CamDavidsonPilon:
Fixed in 0.25.11

PR #1252, opened 2021-04-10
PR for fixing elastic net penalty for cox models

bnaras:
This is a fix for issue #1251 as noted.


CamDavidsonPilon:
Hi @bnaras 

Wow nice catch. Thanks for the PR!

PR #1268, opened 2021-04-29
Fix piecewise hazard prediction

griffiri:
The unit test test_estimation / TestCoxPHFitterPieces.test_predict_hazard_sans_strata fails, where i am comparing the result of predicting hazard for 1 of 5 samples to the 1 sample run by itself. This seems to be because running egrad on a matrix input isn't what we want to do (we just want to differentiate cumulative hazard along time axis). Not sure how to fix it though.

Ignore the change to pre-commit hook for the moment. It's the only way i could seem to commit files. black throws an error message i can't resolve if i leave it there. Can figure this out later...


lifelines/tests/test_estimation.py
```diff
@@ -2873,7 +2873,7 @@ def test_efron_newtons_method(self, data_nus, cph):
         assert np.abs(newton(X, T, E, W, entries)[0] - -0.0335) < 0.0001
 
 
-class TestCoxPHFitterPeices:
+class TestCoxPHFitterPieces:
```
CamDavidsonPilon, line 2876: hehe ty

CamDavidsonPilon:
The code look reasonable, and I can't spot a problem by looking at it. I'm going to pull this locally to see if I can figure it out.

BTW - this specific code, the cumulative hazard function for piecewise-whatever, is very complex (and lacking comments - thanks @CamDavidsonPilon...), and I am surprised you were not scared off by it.

PR #1275, opened 2021-05-22
add BIC_ prop

yildizyasin:
Bayesian Information Criteria (BIC) value is added as model prop: `model.BIC_`

`find_best_parametric_model()` 's `evaluation_lookup` now uses this prop for BIC value.

### DIFF
--- a/lifelines/fitters/__init__.py
+++ b/lifelines/fitters/__init__.py
@@ -309,6 +309,10 @@ class ParametricUnivariateFitter(UnivariateFitter):
     def AIC_(self) -> float:
         return -2 * self.log_likelihood_ + 2 * self._fitted_parameters_.shape[0]
 
+    @property
+    def BIC_(self) -> float:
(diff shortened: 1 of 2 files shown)


lifelines/fitters/__init__.py
```diff
@@ -2244,6 +2249,7 @@ def print_summary(self, decimals=2, style=None, columns=None, **kwargs):
         footers.extend(
             [
                 ("AIC", "{:.{prec}f}".format(self.AIC_, prec=decimals)),
+                ("BIC", "{:.{prec}f}".format(self.BIC_, prec=decimals)),
```
CamDavidsonPilon, line 2252: I like everything, but am unsure about this. I don't want print_summary to be overloaded with metrics / information. Already AIC is there - does having BIC there add much value to the user? What are your thoughts?

CamDavidsonPilon:
Thanks @yildizyasin  - PR looks well done. I have one comment, otherwise LGTM

PR #1276, opened 2021-05-26
Added custom knots parameter for Cox Fitter

JoseLlanes:
I have added a new parameter in CoxPHFitter, named 'knots'. This parameter allows specifying the position of the knots in fitting process, instead of having the library automatically calculating it.

There's also an example file which shows how to use it, by fitting it to one week, a month and a year.

Finally this PR also adds the Pycharm .idea folder to the gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ docs/_build/
 
 # asv builds
 \.asv/
+
+# Pycharm
+\.idea/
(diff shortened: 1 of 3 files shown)


CamDavidsonPilon:
Great contribution, @JoseLlanes! Code looks well done. Can you add this new parameter to the CoxPHFitter's docstring - and some mention that it is mutually exclusive w.r.t. the `n_baseline_knots` param?

JoseLlanes:
Docstring for knots parameter added. How does it look like now ?

CamDavidsonPilon:
Perfect - lgtm! :shipit:

PR #1278, opened 2021-05-26
fix at_risk table bug, and bump version

CamDavidsonPilon:
##### New features
 - `.BIC_` is now present on fitted models.
 - `CoxPHFitter` with spline baseline can accept pre-computed knot locations.
 - Left censoring fitting in KaplanMeierFitter is now "expected". That is, `predict` _always_ predicts the survival function (as does every other model), `confidence_interval_` is _always_ the CI for the survival function (as does every other model), and so on. In summary: the API for estimates doesn't change depending on what your censoring your dataset is.

##### Bug fixes
 - Fixed an annoying bug where at_risk-table label's were not aligning properly when data spanned large ranges. See merging PR for details.
 - Fixed a bug in `find_best_parametric_model` where the wrong BIC value was being computed.
 - Fixed regression bug when using an array as a penalizer in Cox models.

cc @JoseLlanes @yildizyasin your contributions will be released in this version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 ## Changelog
 
+#### 0.25.11 - 2021-05-26
+
+##### New features
+ - `.BIC_` is now present on fitted models.
+ - `CoxPHFitter` with spline baseline can accept pre-computed knot locations.
+
+
+##### Bug fixes
+ - Fixed an annoying bug where at_risk-table label's were not aligning properly when data spanned large ranges. See merging PR for details.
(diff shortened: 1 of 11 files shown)


CamDavidsonPilon:
That annoying issue: 

### Previous:

Screen Shot 2021-05-26 at 7 25 58 PM
Screen Shot 2021-05-26 at 7 32 45 PM



### New:

Screen Shot 2021-05-26 at 7 25 33 PM
Screen Shot 2021-05-26 at 7 32 12 PM

PR #1288, opened 2021-06-09
Fix sorting in interval-censored lifetimes plot

### LINKED ISSUE #1287: ValueError in plot_interval_censored_lifetimes when passing entry dates
The following lines in `plotting.plot_interval_censored_lifetimes(..., entry=None, ....)` triggers a 
`ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().` when the argument `entry` is passed. Furthermore `lower_bound` is wrongly substituted for `entry`.

https://github.com/CamDavidsonPilon/lifelines/blob/71f4e4d659870058c33572e7e582654f96c1fcaa/lifelines/plotting.py#L620-L621

Instead it should be:

```python
        if entry is not None:
            entry = _iloc(entry, ix)
```

Cimbali:
Fixes #1287

### DIFF
--- a/lifelines/plotting.py
+++ b/lifelines/plotting.py
@@ -617,8 +617,8 @@ def plot_interval_censored_lifetimes(
         ix = np.argsort(lower_bound, 0)
         upper_bound = _iloc(upper_bound, ix)
         lower_bound = _iloc(lower_bound, ix)
-        if entry:
-            entry = _iloc(lower_bound, ix)
+        if entry is not None:
+            entry = _iloc(entry, ix)
 
     if entry is None:
         entry = np.zeros(N)


lifelines/plotting.py
```diff
@@ -617,8 +617,8 @@ def plot_interval_censored_lifetimes(
         ix = np.argsort(lower_bound, 0)
         upper_bound = _iloc(upper_bound, ix)
         lower_bound = _iloc(lower_bound, ix)
-        if entry:
-            entry = _iloc(lower_bound, ix)
+        if entry is not None:
+            entry = _iloc(entry, ix)
```
CamDavidsonPilon, line 621: 👍

CamDavidsonPilon:
thanks @Cimbali!

PR #1323, opened 2021-09-15
V0.26.1

CamDavidsonPilon:
##### API Changes
 - `t_0` in `logrank_test` now will not remove data, but will instead censor all subjects that experience the event afterwards.
 - update `status` column in `lifelines.datasets.load_lung` to be more standard coding: 0 is censored, 1 is event.

##### Bug fixes
 - Fix using formulas with `AalenAdditiveFitter.predict_cumulative_hazard`
 - Fix using formulas with `CoxPHFitter.score`


lifelines/utils/__init__.py
```diff
@@ -751,6 +752,9 @@ def k_fold_cross_validation(
     n, _ = df.shape
     df = df.copy()
 
+    if seed is not None:
```
CamDavidsonPilon, line 755: closes #1324

PR #1379, opened 2022-01-07
added kwarg for scatter color to qq_plot

kylejcaron:
When working with hierarchical data, it makes sense to look at separate qq plots for each unit. By having a `scatter_color` argument, qq_plots of multiple units are easier to overlay in the same plot.

### DIFF
--- a/lifelines/plotting.py
+++ b/lifelines/plotting.py
@@ -219,7 +219,7 @@ def rmst_plot(model, model2=None, t=np.inf, ax=None, text_position=None, **plot_
     return ax
 
 
-def qq_plot(model, ax=None, **plot_kwargs):
+def qq_plot(model, ax=None, scatter_color="k", **plot_kwargs):
     """
     Produces a quantile-quantile plot of the empirical CDF against
     the fitted parametric CDF. Large deviances away from the line y=x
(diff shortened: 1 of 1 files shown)


lifelines/plotting.py
```diff
@@ -219,7 +219,7 @@ def rmst_plot(model, model2=None, t=np.inf, ax=None, text_position=None, **plot_
     return ax
 
 
-def qq_plot(model, ax=None, **plot_kwargs):
+def qq_plot(model, ax=None, scatter_color="k", **plot_kwargs):
```
CamDavidsonPilon, line 222: strange: it seems I don't even use `plot_kwargs`. Probably will be removed in a later version.

lifelines/plotting.py
```diff
@@ -291,7 +291,7 @@ def qq_plot(model, ax=None, **plot_kwargs):
 
     max_, min_ = quantiles[COL_EMP].max(), quantiles[COL_EMP].min()
 
-    quantiles.plot.scatter(COL_THEO, COL_EMP, c="none", edgecolor="k", lw=0.5, ax=ax)
+    quantiles.plot.scatter(COL_THEO, COL_EMP, c="none", edgecolor=scatter_color, lw=0.5, ax=ax)
     ax.plot([min_, max_], [min_, max_], c="k", ls=":", lw=1.0)
```
CamDavidsonPilon, line 295: should the color here be the same as `scatter_color`?

CamDavidsonPilon:
Hi @kylejcaron, thanks for the PR. I had a question for you - see comments

lifelines/plotting.py
kylejcaron: Oh woah I didnt notice that either, nice call out

lifelines/plotting.py
kylejcaron: I'd recommend no (I'm assuming you're referring to the `c="k"` argument for ax.plot()?), since that line of code is simply for the diagonal reference line in the QQ plot - for the use case of overlaying qq plots, having different colored diagonals wouldnt make a difference since they all would overlap anyway!

Also sorry this took so long, I'm terrible at noticing github notifications!

CamDavidsonPilon:
lgtm - thanks @kylejcaron!

CamDavidsonPilon:
This be broadly available in the next release

PR #1419, opened 2022-05-24
Print with line break

jseabold:
Updated two of the iteration print statements that weren't to use linebreaks. Otherwise, the iterations get overwritten in a notebook and are jammed together in the terminal.


CamDavidsonPilon:
lgtm, thanks @jseabold

PR #1469, opened 2022-11-01
Fix add_at_risk_counts()

ewulczyn:
Update lambda in agg() to return a scalar, so that agg() returns a pd.Series instead of a pd.DataFrame, as is expected by downstream logic.


CamDavidsonPilon:
I think this makes sense. Can you confirm which version of pandas you are using?

Issue #1512, opened 2023-04-03
Compatibility issue with pandas 2.0.0


DrShushen:
Hi,

There appears to be a problem with `pandas` [`2.0.0`](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html).

Some system info, hopefully useful:
OS: Ubuntu 22.04
Versions:
```txt
lifelines 0.27.4
pandas 2.0.0
numpy 1.23.5
```

Replicating the error:
```python
# reproduce.py

from lifelines.datasets import load_rossi
from lifelines import CoxPHFitter

rossi = load_rossi()
cph = CoxPHFitter().fit(rossi, 'week', 'arrest')

axes = cph.check_assumptions(rossi, show_plots=True)
```
This throws the error:
```txt
Traceback (most recent call last):
  File "./reproduce.py", line 15, in <module>
    fitter.fit(rossi, "week", "arrest")
  File "/<...>/lifelines/utils/__init__.py", line 56, in f
    return function(model, *args, **kwargs)
  File "/<...>/lifelines/fitters/coxph_fitter.py", line 290, in fit
    self._model = self._fit_model(
  File "/<...>/lifelines/fitters/coxph_fitter.py", line 610, in _fit_model
    return self._fit_model_breslow(*args, **kwargs)
  File "/<...>/lifelines/fitters/coxph_fitter.py", line 623, in _fit_model_breslow
    model.fit(*args, **kwargs)
  File "/<...>/lifelines/utils/__init__.py", line 56, in f
    return function(model, *args, **kwargs)
  File "/<...>/lifelines/fitters/coxph_fitter.py", line 1242, in fit
    self._central_values = self._compute_central_values_of_raw_training_data(df, self.strata)
  File "/<...>/lifelines/fitters/__init__.py", line 1286, in _compute_central_values_of_raw_training_data
    described = df.describe(include="all", **describe_kwarg)
TypeError: describe() got an unexpected keyword argument 'datetime_is_numeric'
```

Seems to be the same issue as [this](https://github.com/mlflow/mlflow/pull/7898), fixed as [here](https://github.com/mlflow/mlflow/pull/7898/files).

CamDavidsonPilon:
Thanks @DrShushen for the report! I suspect this isn't the only issue. I'll update soon, as I'd like to test out new Pandas features too

DrShushen:
@CamDavidsonPilon
Great, no problem!

[...]

PR #1516, opened 2023-04-26
fix datetime_is_numeric problem with pandas >= 2.0.0

jonathan-taylor:
Tiny change to fix https://github.com/CamDavidsonPilon/lifelines/issues/1512


CamDavidsonPilon:
that's all that's need! lgtm then!

PR #1528, opened 2023-05-25
ENH use scientific number formatting for delta and newton decrement in Cox fitters

mathurinm:
Hello,
When pushing the solver to convergence, both `delta` and `newton_decrement` go to zero, so the displayed information looks like this when they becomes smaller than 1e-5:
image

With the proposed change a useful value is printed all along convergence:
image


CamDavidsonPilon:
Love it!

CamDavidsonPilon:
Thanks @mathurinm

PR #1534, opened 2023-06-14
expose r_precision parameter in Newton raphson solvers, correct typos

mathurinm:
The stopping criterion of the Newton-Raphson method is mostly controlled by the parameter `precision`, but there is also a stopping criterion based on the relative improvement of the log-likelihood that so far used a hardcoded value of 1e-9. This PR exposes this parameter in the two NR solvers, with default value 1e-9 so as not to break existing code.

It also harmonizes stopping criteria between time varying and non time varying Cox Fitter

Finally it corrects the spelling of Raphson in docstrings, warning messages and comments

ping @Badr-Moufad

### DIFF
--- a/lifelines/fitters/coxph_fitter.py
+++ b/lifelines/fitters/coxph_fitter.py
@@ -242,7 +242,7 @@ class CoxPHFitter(RegressionFitter, ProportionalHazardMixin):
             algorithm. Default is the zero vector.
 
         fit_options: dict, optional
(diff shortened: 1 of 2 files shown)


lifelines/fitters/coxph_fitter.py
```diff
@@ -1594,6 +1597,7 @@ def _newton_raphson_for_efron_model(
             step_size = step_sizer.update(norm_delta).next()
 
         if show_progress and success:
+            import ipdb; ipdb.set_trace()
```
Badr-MOUFAD, line 1600: it has slipped through your keyboard

mathurinm: thanks, fixed!

PR #1547, opened 2023-07-20
Fixed example and reproducibility for Log-likelihood

### LINKED ISSUE #1545: Log likelihood needs to be maximized or minimized?
Hi, first of all, thank you very much for this library! :100: 

I'm reading the [Log Likelihood documentation](https://lifelines.readthedocs.io/en/latest/Survival%20Regression.html?highlight=likelihood#log-likelihood) and it says:

> The in-sample log-likelihood is available under log_likelihood_ of any regression model. For out-of-sample data, the score() method (available on all regression models) can be used. This returns the average evaluation of the out-of-sample log-likelihood. **We want to maximize this**.

But I'm looking that, in the example below, there is a comment `better model` that corresponds to the **lowest** (not highest) value:

```pyhon
from lifelines import CoxPHFitter
from lifelines.datasets import load_rossi

rossi = load_rossi().sample(frac=1.0)
train_rossi = rossi.iloc[:400]
test_rossi = rossi.iloc[400:]

cph_l2 = CoxPHFitter(penalizer=0.1, l1_ratio=0.).fit(train_rossi, 'week', 'arrest')
cph_l1 = CoxPHFitter(penalizer=0.1, l1_ratio=1.).fit(train_rossi, 'week', 'arrest')

res_l2 = cph_l2.score(test_rossi)
res_l1 = cph_l1.score(test_rossi)

print(res_l2)
print(res_l1) # better model

print(res_l2 < res_l1)  # Prints False, so the model with l1_ratio=1 is lower, it's not the maximum
```

Is this an error in the code example or should the log-likelihood be minimized instead of maximized?

I really appreciate any help you can provide

Genarito:
Fixes #1545: a model with the highest Log-Likelihood is the best one. Also, `random_state` prevents Pandas random sampling from generating a counterexample to what the documentation states (sometimes the "best model" could be the lower score)

### DIFF
--- a/docs/Survival Regression.rst
+++ b/docs/Survival Regression.rst
@@ -1134,15 +1134,15 @@ In this author's opinion, the best way to measure predictive performance is eval
     from lifelines import CoxPHFitter
     from lifelines.datasets import load_rossi
 
-    rossi = load_rossi().sample(frac=1.0)
+    rossi = load_rossi().sample(frac=1.0, random_state=25)  # ensures the reproducibility of the example
     train_rossi = rossi.iloc[:400]
     test_rossi = rossi.iloc[400:]
 
-    cph_l2 = CoxPHFitter(penalizer=0.1, l1_ratio=0.).fit(train_rossi, 'week', 'arrest')
     cph_l1 = CoxPHFitter(penalizer=0.1, l1_ratio=1.).fit(train_rossi, 'week', 'arrest')
+    cph_l2 = CoxPHFitter(penalizer=0.1, l1_ratio=0.).fit(train_rossi, 'week', 'arrest')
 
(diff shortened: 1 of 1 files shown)


docs/Survival Regression.rst
```diff
@@ -1134,15 +1134,15 @@ In this author's opinion, the best way to measure predictive performance is eval
     from lifelines import CoxPHFitter
     from lifelines.datasets import load_rossi
 
-    rossi = load_rossi().sample(frac=1.0)
+    rossi = load_rossi().sample(frac=1.0, random_state=25)  # ensures the reproducibility of the example
```
CamDavidsonPilon, line 1137: nice addition!

Issue #1574, opened 2023-11-13
Fix confidence_interval_cumulative_density_


TheooJ:
Hi, 

I noticed that the confidence intervals for the cumulative density in `KaplanMeierFitter` and `BreslowFlemingHarringtonFitter` were inverted because of the subtraction `1 - self.confidence_interval_`, which leads the lower CI being greater than the upper CI

PR #1575, opened 2023-11-13
switch confidence interval values

TheooJ:
Closing #1574

I switched the values in the two columns, and kept the ordering lower, upper


CamDavidsonPilon:
Right, thanks makes sense. Thanks @TheooJ!

PR #1591, opened 2024-01-15
Fix: Typos

FBruzzesi:
# Description

While reading the docs I found a few typos.
Then I decided to take a systematic approach and fix those I could find.

Some are in code variables which are not exposed to the end user so I felt free to change as well.


CamDavidsonPilon:
Wow you found a lot of errors! Thanks!

PR #1596, opened 2024-02-14
address pandas deprecation warning

mathematicalmichael:
```
.../lib/python3.10/site-packages/lifelines/fitters/cox_time_varying_fitter.py:819: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value '0.6058777002252564' has dtype incompatible with int64, please explicitly cast to a compatible dtype first.
```
this warning goes away with this change.
`pandas==2.1.4`

### DIFF
--- a/lifelines/fitters/cox_time_varying_fitter.py
+++ b/lifelines/fitters/cox_time_varying_fitter.py
@@ -801,7 +801,7 @@ See https://stats.stackexchange.com/questions/11109/how-to-deal-with-perfect-sep
             hazards = self.predict_partial_hazard(tv_data).values
 
         unique_death_times = np.unique(stop[events.values])
-        baseline_hazard_ = pd.DataFrame(np.zeros_like(unique_death_times), index=unique_death_times, columns=["baseline hazard"])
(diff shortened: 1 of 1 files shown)


lifelines/fitters/cox_time_varying_fitter.py
```diff
@@ -801,7 +801,7 @@ def _compute_cumulative_baseline_hazard(self, tv_data, events, start, stop, weig
             hazards = self.predict_partial_hazard(tv_data).values
 
         unique_death_times = np.unique(stop[events.values])
-        baseline_hazard_ = pd.DataFrame(np.zeros_like(unique_death_times), index=unique_death_times, columns=["baseline hazard"])
+        baseline_hazard_ = pd.DataFrame(np.zeros_like(unique_death_times).astype(float), index=unique_death_times, columns=["baseline hazard"])
```
CamDavidsonPilon, line 804: The warning posted says `Value '0.6058777002252564' has dtype incompatible with int64`, but this is the `zeros_like` function - which should produces `0`. I'm confused.

mathematicalmichael: Yeah, because it (a couple lines down) tries to assign something to this column.

PR #1637, opened 2024-09-17
Survival function for Cox Time Varying model

rvdinter:
I've written a method for getting the survival function until timestep $t$ where $0<t<\tau$. It will not show $S(t)$ for all the way to time of failure $\tau$, as you do not know the covariates at that moment. Adding this function supports with calculation of $\int_0^t S(t)$, needed for determining the optimal replacement time.

### DIFF
--- a/lifelines/fitters/cox_time_varying_fitter.py
+++ b/lifelines/fitters/cox_time_varying_fitter.py
@@ -295,6 +295,15 @@ class CoxTimeVaryingFitter(SemiParametricRegressionFitter, ProportionalHazardMix
             index=self.params_.index,
         )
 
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
Neat, thanks @rvdinter! Are you comfortable writing a test for this too? If so, it would go in `test_estimation.py` under the `TestCoxTimeVaryingFitter` class.

rvdinter:
@CamDavidsonPilon definitely! I will work on it.

rvdinter:
@CamDavidsonPilon just added the tests, let me know if I need to change anything!

rvdinter:
@CamDavidsonPilon could you review the PR?

PR #1660, opened 2025-07-05
Optimize variance calculation at AalenJohansenFitter with prefix sum approach

h-b-k-nishi:
While using the library, I noticed that `AalenJohansenFitter.fit()` is very slow. It appears that the variance calculation, enabled by default, constitutes the primary bottleneck, due to its time complexity $O(N ^ 2)$ where $N$ is the number of events. The method iterates through $O(N)$ rows, spending $O(N)$ time to calculate variance at each step.

According to the codebase and comments, the variance for the $r$-th row equals to the sum of three terms `first_term`, `second_term`, and $\\{(-2) \times$ `third_term` $\\}$, where

- `first_term` $= \sum_{i = 1}^{r} \frac{(F_j(t) - F_j(t_i))^2 \times d}{n \times(n - d)}$
- `second_term` $= \sum_{i = 1}^{r} \frac{S^2(t_i - 1) \times d \times (n - d)}{n^3}$
- `third_term` $= \sum_{i = 1}^{r} \frac{(F_j(t) - F_j(t_i)) \times S(t_i - 1) \times d}{n^2}$

Here, $n$ and $d$ are values that depend on the index $i$.

Each of the three terms can be regarded as a quadratic, constant, and linear function respectively in terms of $F_j(t)$. For example, the `first_term` for the $r$-th row can be represented as $P_r(F_j(t))$, where $P_r(x) = \sum_{i = 1}^{r} \frac{(x - F_j(t_i))^2 \times d}{n \times(n-d)}$.

Since $P_r(x) = P_{r-1}(x) + \frac{(x - F_j(t_r))^2 \times d}{n \times (n-d)}$ for $r = 1, 2, \dots, N$ with $P_0(x) = 0$, we can use a prefix sum algorithm to iteratively update $P_r(x)$ using a data structure that represents a quadratic function and supports:

- Adding a term $\frac{(x - F_j(t_r))^2 \times d}{n \times (n-d)}$, and
- Evaluating $P(x)$ at the given value of $x$.

I will omit details of the data structure implementation here, but the main idea is to store the quadratic, linear, and constant terms of the function.

By applying this method, the performance improved by more than $\text{40x}$ as $N$ approaches $20,000$--no longer requiring a wait of several seconds or even minutes!

<p float="left">
Screenshot 2025-07-05 at 14 08 56
Screenshot 2025-07-05 at 14 09 12
</P>

We could further optimize performance by using `cumsum()` to evaluate the functions directly and avoid relying on `iterrows()`, but for the sake of maintainability, I chose to continue using `iterrows()`.

### DIFF
--- a/lifelines/fitters/aalen_johansen_fitter.py
+++ b/lifelines/fitters/aalen_johansen_fitter.py
@@ -5,7 +5,7 @@ import pandas as pd
 import warnings
 
 from lifelines.fitters import NonParametricUnivariateFitter
-from lifelines.utils import _preprocess_inputs, inv_normal_cdf, CensoringType, coalesce
+from lifelines.utils import _preprocess_inputs, inv_normal_cdf, CensoringType, coalesce, LinearAccumulator, QuadraticAccumulator
 from lifelines import KaplanMeierFitter
 from lifelines.plotting import _plot_estimate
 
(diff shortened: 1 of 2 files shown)


h-b-k-nishi:
@CamDavidsonPilon May I kindly ask for your review? This update was made in response to a niche use case--using the library for a dashboard that displays statistical summaries of hundreds of thousands of data points. However, I believe this speed-up will be beneficial more broadly.

h-b-k-nishi:
I added unit tests for the accumulators to ensure correctness and prevent potential errors 🙇

h-b-k-nishi:
@CamDavidsonPilon If any documentation or code adjustments are needed, please feel free to let me know--I'd be more than happy to contribute further. In some use cases, the number of data points can actually grow to between $100,000$ and $1,000,000$, and I believe the performance improvement would be even more significant in such scenarios.

CamDavidsonPilon:
Hey, this is very cool. Well done! I'll merge test locally too.

PR #1675, opened 2026-03-06
PERF: Optimise to_long_format by eliminating intermediate DataFrame allocations

matheusft:
## Description

This PR improves the performance of `lifelines.utils.to_long_format`. 

The previous implementation relied on method chaining (`df.assign(...).drop(...)`), which inadvertently triggered two full DataFrame allocations in memory:
1. `assign` created the first copy to append the new columns.
2. `drop` created a second copy to remove the `duration_col`.

By refactoring this to perform a single explicit `.copy()` followed by fast, in-place operations (`.pop()` and direct assignment), we effectively halve the memory overhead and noticeably speed up execution on large DataFrames. 

This change introduces no breaking changes and is fully backward-compatible.

## Type of change
- [x] Performance improvement (non-breaking change that speeds up existing functionality)
- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/lifelines/utils/__init__.py
+++ b/lifelines/utils/__init__.py
@@ -1323,7 +1323,14 @@ def to_long_format(df, duration_col) -> pd.DataFrame:
     to_episodic_format
     add_covariate_to_timeline
     """
-    return df.assign(start=0, stop=lambda s: s[duration_col]).drop(duration_col, axis=1)
+    # Create a single copy to avoid modifying the original dataframe
+    long_form_df = df.copy()
+    
+    # pop() removes the column in-place and returns it, avoiding the need for .drop()
+    long_form_df['stop'] = long_form_df.pop(duration_col)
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
👍 I love perf wins like this! Have any numbers to share?

matheusft:
<p>Here are the performance numbers comparing the existing <code>new_to_long_format</code> with the optimised <code>original_to_long_format</code>, using a df (203,769 rows × 43 columns, no missing data, ~29.5 MB):</p>
<h3>⏱️ <strong>Performance</strong></h3>


Function | Median time (sec) | Mean (sec) | Speed vs. slow
-- | -- | -- | -- 
new_to_long_format | 0.0063 | 0.00635 | ~1.8× faster
original_to_long_format | 0.0112 | 0.01108 | baseline

Numbers obtained after running each function 10x

